PR #20, opened 2022-09-29
add GEE regression verbose

oaclavijo10:


cluster_experiments/experiment_analysis.py
```diff
@@ -121,7 +122,7 @@ def __init__(
         self.fam = sm.families.Gaussian()
         self.va = sm.cov_struct.Exchangeable()
 
-    def analysis_pvalue(self, df: pd.DataFrame) -> float:
+    def analysis_pvalue(self, df: pd.DataFrame, verbose=False) -> float:
```
david26694, line 125: Can you add type hints?

And can you add it to the docstring of the method?

david26694 (changes requested):

david26694 (changes requested):
Have you got black running? If not, can you do it?

Apparently, the CI is failing

image

PR #24, opened 2022-10-03
add stratified splitters

david26694:


ludovico-lanni (approved):
looks good to me, I agree on using the assign_treatment_df method as the first choice for developing a custom splitter! 👍

PR #80, opened 2023-01-30
implement paired t test

Gabrielcidral1:
Implement paired t test, based on ttest_rel (scipy.stats).

Strata is the column that will decide the comparison and the columns are generated from treatment_col. 

Minor comment:
In test_analysis.test_paired_ttest I am generating some fake data because it's specific to this test. If relevant I can pass it to a generic function so it's reusable.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -17,3 +17,8 @@ repos:
     hooks:
       - id: black
         language_version: python3.8
+-   repo: https://github.com/timothycrosley/isort
+    rev: 5.6.4
+    hooks:
+    -   id: isort
+        args: ["--profile", "black"]
--- a/cluster_experiments/experiment_analysis.py
+++ b/cluster_experiments/experiment_analysis.py
@@ -1,6 +1,6 @@
 from abc import ABC, abstractmethod
 from typing import List, Optional
-
+import warnings
 import pandas as pd
 import statsmodels.api as sm
 from pandas.api.types import is_numeric_dtype
@@ -310,7 +310,7 @@ class PairedTTestClusteredAnalysis(ExperimentAnalysis):
         target_col: name of the column containing the variable to measure
         treatment_col: name of the column containing the treatment variable
         treatment: name of the treatment to use as the treated group
-        comparison_col: column to generate 2 groups (a and b)
+        strata: column to generate 2 groups (a and b)
 
     Usage:
 
@@ -324,9 +324,9 @@ class PairedTTestClusteredAnalysis(ExperimentAnalysis):
         'cluster': [1, 2, 3, 4, 1, 2, 3, 4],
     })
 
-    TTestClusteredAnalysis(
-        cluster_cols=['cluster'],
-        target_col='x',
+    PairedTTestClusteredAnalysis(
+        strata=df['cluster'],
+        target_col=df['x'],
     ).get_pvalue(df)
     ```
     """
@@ -334,12 +334,12 @@ class PairedTTestClusteredAnalysis(ExperimentAnalysis):
     def __init__(
         self,
         cluster_cols: List[str],
+        strata_cols: List[str] = None,
         target_col: str = "target",
         treatment_col: str = "treatment",
         treatment: str = "B",
-        comparison_col: str = "",
     ):
-        self.comparison_col = comparison_col
+        self.strata_cols = cluster_cols if strata_cols is None else strata_cols
         self.target_col = target_col
         self.treatment = treatment
         self.treatment_col = treatment_col
@@ -356,15 +356,21 @@ class PairedTTestClusteredAnalysis(ExperimentAnalysis):
             self.cluster_cols + [self.treatment_col], as_index=False
         )[self.target_col].mean()
 
-        assert (
-            df_grouped["comparison_col"].nunique() == 2
-        ), "there should be only 2 values for paired t test"
+        if (
+            df_grouped[self.treatment_col].value_counts()[0]
+            != df_grouped[self.treatment_col].value_counts()[1]
+        ):
+            warnings.warn(
+                "groups don't have same number of observations"
+            )  # todo raise count as well?
 
         df_pivot = df_grouped.pivot_table(
             columns=self.treatment_col,
-            index=self.comparison_col,
+            index=self.strata_cols,
             values=self.target_col,
         )
+
+        assert df_pivot.isna().sum().sum() == 0, "missing data from some cluster"
 
         t_test_results = ttest_rel(df_pivot.iloc[:, 0], df_pivot.iloc[:, 1])
         return t_test_results.pvalue
@@ -377,7 +383,7 @@ class PairedTTestClusteredAnalysis(ExperimentAnalysis):
             target_col=config.target_col,
             treatment_col=config.treatment_col,
             treatment=config.treatment,
-            comparison_col=config.comparison_col,
+            strata=config.strata_cols,
         )
 
 
--- a/cluster_experiments/experiment_analysis.py
+++ b/cluster_experiments/experiment_analysis.py
@@ -1,6 +1,7 @@
 import logging
 from abc import ABC, abstractmethod
 from typing import List, Optional
+
 import pandas as pd
 import statsmodels.api as sm
 from pandas.api.types import is_numeric_dtype
@@ -326,6 +327,7 @@ class PairedTTestClusteredAnalysis(ExperimentAnalysis):
     })
 
     PairedTTestClusteredAnalysis(
+        cluster_cols=df['cluster'],
         strata=df['cluster'],
         target_col=df['x'],
     ).get_pvalue(df)
--- a/cluster_experiments/experiment_analysis.py
+++ b/cluster_experiments/experiment_analysis.py
@@ -77,7 +77,7 @@ class ExperimentAnalysis(ABC):
                 f"Outcome column {self.target_col} should be numeric and not {df[self.target_col].dtype}"
             )
 
-    def get_pvalue(self, df: pd.DataFrame) -> float:
+    def get_pvalue(self, df: pd.DataFrame, verbose: bool) -> float:
         """Returns the p-value of the analysis
 
         Arguments:
@@ -86,7 +86,7 @@ class ExperimentAnalysis(ABC):
         df = df.copy()
         df = self._create_binary_treatment(df)
         self._data_checks(df=df)
-        return self.analysis_pvalue(df)
+        return self.analysis_pvalue(df, verbose=verbose)
 
     @classmethod
     def from_config(cls, config):
@@ -347,7 +347,7 @@ class PairedTTestClusteredAnalysis(ExperimentAnalysis):
         self.treatment_col = treatment_col
         self.cluster_cols = cluster_cols
 
-    def _preprocessing(self, df: pd.DataFrame) -> pd.DataFrame:
+    def _preprocessing(self, df: pd.DataFrame, verbose: bool = False) -> pd.DataFrame:
         df_grouped = df.groupby(
             self.cluster_cols + [self.treatment_col], as_index=False
         )[self.target_col].mean()
@@ -370,10 +370,13 @@ class PairedTTestClusteredAnalysis(ExperimentAnalysis):
             values=self.target_col,
         )
 
-        if df_pivot.isna().sum().sum() == 0:
+        if df_pivot.isna().sum().sum() > 0:
             logging.warning(
-                f"There are missing data from some clusters, removing them: {df_pivot[df_pivot.isna().any(axis=1)].to_dict()}"
+                f"There are missing pairs for some clusters, removing the lonely ones: {df_pivot[df_pivot.isna().any(axis=1)].to_dict()}"
             )
+
+        if verbose:
+            print(f"performing paired t test in this data \n {df_pivot} \n")
 
         df_pivot = df_pivot.dropna()
 
@@ -383,13 +386,10 @@ class PairedTTestClusteredAnalysis(ExperimentAnalysis):
         """Returns the p-value of the analysis
         Arguments:
             df: dataframe containing the data to analyze
-            verbose (Optional): bool, prints the regression summary if True
+            verbose (Optional): bool, prints the extra info if True
         """
 
         df_pivot = self._preprocessing(df=df)
-
-        if verbose:
-            print(f"performing paired t test in this data \n {df_pivot} \n")
 
         t_test_results = ttest_rel(df_pivot.iloc[:, 0], df_pivot.iloc[:, 1])
 
--- a/cluster_experiments/experiment_analysis.py
+++ b/cluster_experiments/experiment_analysis.py
@@ -369,6 +369,14 @@ class PairedTTestClusteredAnalysis(ExperimentAnalysis):
             index=self.strata_cols,
             values=self.target_col,
         )
+
+        if df_pivot.isna().sum().sum() == 0:
+            logging.warning(
+                f"There are missing data from some clusters, removing them: {df_pivot[df_pivot.isna().any(axis=1)].to_dict()}"
+            )
+
+        df_pivot = df_pivot.dropna()
+
         return df_pivot
 
     def analysis_pvalue(self, df: pd.DataFrame, verbose: bool = False) -> float:
@@ -379,8 +387,6 @@ class PairedTTestClusteredAnalysis(ExperimentAnalysis):
         """
 
         df_pivot = self.preprocessing(df=df)
-
-        assert df_pivot.isna().sum().sum() == 0, "missing data from some cluster"
 
         if verbose:
             f"performing paired t test in this data \n {df_pivot}"
--- a/cluster_experiments/experiment_analysis.py
+++ b/cluster_experiments/experiment_analysis.py
@@ -356,17 +356,17 @@ class PairedTTestClusteredAnalysis(ExperimentAnalysis):
             self.cluster_cols + [self.treatment_col], as_index=False
         )[self.target_col].mean()
 
+        assert (
+            df_grouped["comparison_col"].nunique() == 2
+        ), "there should be only 2 values for paired t test"
+
         df_pivot = df_grouped.pivot_table(
             columns=self.treatment_col,
             index=self.comparison_col,
             values=self.target_col,
         )
 
-        assert (
-            df.pivot.shape[1] == 2
-        ), "There should be only 2 columns (control and treatment)"
-
-        t_test_results = ttest_rel(df_pivot[0], df_pivot[1])
+        t_test_results = ttest_rel(df_pivot.iloc[:, 0], df_pivot.iloc[:, 1])
         return t_test_results.pvalue
 
     @classmethod
--- a/cluster_experiments/experiment_analysis.py
+++ b/cluster_experiments/experiment_analysis.py
@@ -310,6 +310,7 @@ class PairedTTestClusteredAnalysis(ExperimentAnalysis):
         target_col: name of the column containing the variable to measure
         treatment_col: name of the column containing the treatment variable
         treatment: name of the treatment to use as the treated group
+        comparison_col: column to generate 2 groups (a and b)
 
     Usage:
 
@@ -336,7 +337,7 @@ class PairedTTestClusteredAnalysis(ExperimentAnalysis):
         target_col: str = "target",
         treatment_col: str = "treatment",
         treatment: str = "B",
-        comparison_col: str or None = None,
+        comparison_col: str = "",
     ):
         self.comparison_col = comparison_col
         self.target_col = target_col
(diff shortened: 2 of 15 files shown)


FranDeLio (approved):

cluster_experiments/experiment_analysis.py
```diff
@@ -301,6 +301,92 @@ def from_config(cls, config):
...
+            df_grouped[self.treatment_col].value_counts()[0]
+            != df_grouped[self.treatment_col].value_counts()[1]
+        ):
+            warnings.warn(
+                "groups don't have same number of observations"
+            )  # todo raise count as well?
+
+        df_pivot = df_grouped.pivot_table(
+            columns=self.treatment_col,
+            index=self.strata_cols,
```
ludovico-lanni, line 369: in this case, you are using strata_cols (or cluster_cols) as the identifier of the pair, right? What if I have cities as strata_cols and couriers as cluster_cols ? In that case, should the identifier of the pair be cluster_cols + strata_cols in the most generic scenario?

ludovico-lanni (changes requested):
added a comment :)

cluster_experiments/experiment_analysis.py
```diff
@@ -1,10 +1,10 @@
 from abc import ABC, abstractmethod
 from typing import List, Optional
-
+import warnings
```
david26694, line 3: have you run pre-commit install? I believe this is not following the isort behaviour

cluster_experiments/experiment_analysis.py
```diff
@@ -301,6 +301,92 @@ def from_config(cls, config):
...
+        strata=df['cluster'],
+        target_col=df['x'],
+    ).get_pvalue(df)
+    ```
+    """
+
+    def __init__(
+        self,
+        cluster_cols: List[str],
+        strata_cols: List[str] = None,
```
david26694, line 337: doesn't follow type hint, if you want it to be None, should be Optional[List[str]]. I would go for not allowing it to be none anyway

cluster_experiments/experiment_analysis.py
```diff
@@ -301,6 +301,92 @@ def from_config(cls, config):
...
+        cluster_cols: list of columns to use as clusters
+        target_col: name of the column containing the variable to measure
+        treatment_col: name of the column containing the treatment variable
+        treatment: name of the treatment to use as the treated group
+        strata: column to generate 2 groups (a and b)
+
+    Usage:
+
+    ```python
+    from cluster_experiments.experiment_analysis import TTestClusteredAnalysis
```
david26694, line 318: you're not importing paired ttest (I think the unit tests will fail because of this)

cluster_experiments/experiment_analysis.py
```diff
@@ -301,6 +301,92 @@ def from_config(cls, config):
...
+class PairedTTestClusteredAnalysis(ExperimentAnalysis):
+    """
+    Class to run paired T-test analysis on aggregated data
+
+    Arguments:
+        cluster_cols: list of columns to use as clusters
+        target_col: name of the column containing the variable to measure
+        treatment_col: name of the column containing the treatment variable
+        treatment: name of the treatment to use as the treated group
+        strata: column to generate 2 groups (a and b)
```
david26694, line 313: strata_cols

cluster_experiments/power_config.py
```diff
@@ -76,6 +77,7 @@ class PowerConfig:
 
     # optional mappings
     cupac_model: str = ""
+    comparison_col: str = ""
```
david26694, line 80: drop this

cluster_experiments/experiment_analysis.py
```diff
@@ -301,6 +301,92 @@ def from_config(cls, config):
...
+        self.cluster_cols = cluster_cols
+
+    def analysis_pvalue(self, df: pd.DataFrame, verbose: bool = False) -> float:
+        """Returns the p-value of the analysis
+        Arguments:
+            df: dataframe containing the data to analyze
+            verbose (Optional): bool, prints the regression summary if True
+        """
+
+        df_grouped = df.groupby(
```
david26694, line 355: I would extract all the preprocessing to a method and unit-test the method that does the preprocessing

cluster_experiments/experiment_analysis.py
```diff
@@ -301,6 +301,92 @@ def from_config(cls, config):
...
+            self.cluster_cols + [self.treatment_col], as_index=False
+        )[self.target_col].mean()
+
+        if (
+            df_grouped[self.treatment_col].value_counts()[0]
+            != df_grouped[self.treatment_col].value_counts()[1]
+        ):
+            warnings.warn(
```
david26694, line 363: in other parts of the library I've raised warnings with logging

cluster_experiments/experiment_analysis.py
```diff
@@ -301,6 +301,92 @@ def from_config(cls, config):
...
+                "groups don't have same number of observations"
+            )  # todo raise count as well?
```
david26694, line 365: yes!

david26694 (changes requested):
left some comments, unit tests are failing

cluster_experiments/experiment_analysis.py
Gabrielcidral1: hm not sure. I think we should concatenate them because the strata can be different than the cluster_cols in case you  want to compare different things. In your example, if I leave only cities in strata, every row will be a city and it will average all couriers. If you have cluster_cols + strata_cols in strata every row will be compararing the courier level, and  each courier will be either in A or B, right?  For example :

`
def test_paired_ttest():
    "This test make sure that pivot table works as expected (inside get_pvalue) and that paired t test returns a possible value"

    analysis_df = pd.DataFrame(
        {
            "cluster": ["ES"] * 4 + ["IT"] * 4 + ["PL"] * 4 + ["RO"] * 4,
            "date": ["2022-01-01", "2022-01-02", "2022-01-03", "2022-01-04"] * 4,
            "treatment": ["A", "B"] * 8,
            "target": [0.01] * 16,
        }
    )
    # Changing just one observation so we have a p value
    analysis_df.loc[1, "target"] = 0.001

    analyser = PairedTTestClusteredAnalysis(
        cluster_cols=["cluster", "date"], strata_cols=["cluster", "date"]
    )  # todo test passing no strata, test passing many strata

    assert 1 >= analyser.get_pvalue(analysis_df) >= 0, "p value is null or inf"`

Will return:
image

Make sense? Maybe it doesn't 😆

cluster_experiments/experiment_analysis.py
Gabrielcidral1: I don't see isort in pre commit 
image

cluster_experiments/experiment_analysis.py
Gabrielcidral1: So is it preferable to have a mandatory strata_cols?

cluster_experiments/experiment_analysis.py
ludovico-lanni: hmm, I think we should clarify the meaning of strata_cols then. It's not fully clear to me what role does this column has.. My assumption is that cluster_cols and strata_cols are independent one from each other. What I see in the code is that:
- cluster_cols manages the level of pre-aggregation of the metric -> OK
- strata_cols manages a further level of aggregation of the metric, by deciding the level of granularity of the pivot table that puts paired items one next to each other, after having already grouped the rows at the cluster-treatment level. -> Here I'm a bit confused.

Where would I use strata_cols that is different from cluster_cols, and how, in a practical example?

cluster_experiments/experiment_analysis.py
david26694: I think so, the analysis method will fail otherwise, right?

cluster_experiments/experiment_analysis.py
david26694: ah damn, perhaps it is my linter, will investigate

david26694: indeed I have it on my linter, could you add it in here?

david26694 (changes requested):

cluster_experiments/experiment_analysis.py
Gabrielcidral1: Thanks Ludo for comments and David for summarising it. I think the final conclusion is that strata_cols should always be either a subset or equals to cluster_cols

cluster_experiments/experiment_analysis.py
Gabrielcidral1: sure, done

cluster_experiments/experiment_analysis.py
Gabrielcidral1: good idea, done!

.pre-commit-config.yaml
```diff
@@ -17,3 +17,8 @@ repos:
     hooks:
       - id: black
         language_version: python3.8
+-   repo: https://github.com/timothycrosley/isort
+    rev: 5.6.4
+    hooks:
+    -   id: isort
+        args: ["--profile", "black"]
```
Gabrielcidral1, line 24: @david26694 I copied this from ds analytics

Gabrielcidral1: Should I run isort in the entire repo in another pr?

cluster_experiments/experiment_analysis.py
Gabrielcidral1: I disagree @ludovico-lanni. In switchbacks for example you always want to cluster by date, but you will add to the strata cols

ludovico-lanni: ahhh, ok now I get the subset part, I misunderstood it! Ok I see what you mean, I think this serves a specific use case like the one David described.. makes sense to me! Sorry for the late catch 🆙

ludovico-lanni (approved):

.pre-commit-config.yaml
david26694: yeah pls

cluster_experiments/experiment_analysis.py
```diff
@@ -301,6 +302,112 @@ def from_config(cls, config):
...
+        target_col: name of the column containing the variable to measure
+        treatment_col: name of the column containing the treatment variable
+        treatment: name of the treatment to use as the treated group
+        strata_cols: list of index columns for paired t test. Should be a subset or equal to cluster_cols
+
+    Usage:
+
+    ```python
+    from cluster_experiments.experiment_analysis import PairedTTestClusteredAnalysis
+    from scipy.stats import  ttest_rel
```
david26694, line 320: remove this line

cluster_experiments/experiment_analysis.py
```diff
@@ -301,6 +302,112 @@ def from_config(cls, config):
...
+
+    df = pd.DataFrame({
+        'x': [1, 2, 3, 4, 0, 0, 1, 1],
+        'treatment': ["A", "B", "A", "B"] * 2,
+        'cluster': [1, 2, 3, 4, 1, 2, 3, 4],
+    })
+
+    PairedTTestClusteredAnalysis(
+        cluster_cols=df['cluster'],
+        strata=df['cluster'],
```
david26694, line 331: strata_cols

cluster_experiments/experiment_analysis.py
```diff
@@ -301,6 +302,112 @@ def from_config(cls, config):
...
+    import pandas as pd
+
+    df = pd.DataFrame({
+        'x': [1, 2, 3, 4, 0, 0, 1, 1],
+        'treatment': ["A", "B", "A", "B"] * 2,
+        'cluster': [1, 2, 3, 4, 1, 2, 3, 4],
+    })
+
+    PairedTTestClusteredAnalysis(
+        cluster_cols=df['cluster'],
```
david26694, line 330: this should be 'cluster', not df['cluster']. I thought this was supposed to fail in unit tests, have you run pytest?

cluster_experiments/experiment_analysis.py
```diff
@@ -301,6 +302,112 @@ def from_config(cls, config):
...
+        treatment_col: str = "treatment",
+        treatment: str = "B",
+    ):
+        self.strata_cols = cluster_cols if strata_cols is None else strata_cols
+        self.target_col = target_col
+        self.treatment = treatment
+        self.treatment_col = treatment_col
+        self.cluster_cols = cluster_cols
+
+    def preprocessing(self, df):
```
david26694, line 351: Rename to _preprocessing?

cluster_experiments/experiment_analysis.py
```diff
@@ -301,6 +302,112 @@ def from_config(cls, config):
...
+
+    def __init__(
+        self,
+        cluster_cols: List[str],
+        strata_cols: List[str],
+        target_col: str = "target",
+        treatment_col: str = "treatment",
+        treatment: str = "B",
+    ):
+        self.strata_cols = cluster_cols if strata_cols is None else strata_cols
```
david26694, line 345: strata_cols should not be none, why not do self.strata_cols = strata_cols?

cluster_experiments/experiment_analysis.py
```diff
@@ -301,6 +302,112 @@ def from_config(cls, config):
...
+        df_grouped = df.groupby(
+            self.cluster_cols + [self.treatment_col], as_index=False
+        )[self.target_col].mean()
+
+        n_control = df_grouped[self.treatment_col].value_counts()[0]
+        n_treatment = df_grouped[self.treatment_col].value_counts()[1]
+
+        if n_control != n_treatment:
+            logging.warning(
+                f"groups don't have same number of observations, {n_treatment =} and  {n_treatment =}"
```
david26694, line 361: n_control, not n_treatment

cluster_experiments/experiment_analysis.py
```diff
@@ -301,6 +302,112 @@ def from_config(cls, config):
...
+            )
+
+        #         set(self.strata_cols if len ) <= set(self.cluster_cols)
```
david26694, line 364: remove

tests/examples.py
```diff
@@ -51,3 +50,20 @@ def generate_non_clustered_data(N, n_users):
...
+
+def generate_clustered_data():
+    analysis_df = pd.DataFrame(
+        {
+            "country_code": ["ES"] * 4 + ["IT"] * 4 + ["PL"] * 4 + ["RO"] * 4,
+            "city_code": ["BCN", "BCN", "MAD", "BCN"]
+            + ["NAP"] * 4
+            + ["WAW"] * 4
+            + ["BUC"] * 4,
+            "courier_id": [1, 1, 2, 1, 3, 4, 5, 6, 7, 8, 8, 8, 9, 9, 9, 9],
```
david26694, line 63: stupid comment: could we change to user_id?

cluster_experiments/experiment_analysis.py
```diff
@@ -301,6 +302,112 @@ def from_config(cls, config):
...
+    def analysis_pvalue(self, df: pd.DataFrame, verbose: bool = False) -> float:
+        """Returns the p-value of the analysis
+        Arguments:
+            df: dataframe containing the data to analyze
+            verbose (Optional): bool, prints the regression summary if True
+        """
+
+        df_pivot = self.preprocessing(df=df)
+
+        if verbose:
```
david26694, line 393: do you think it makes sense to print the t-test results if verbose?

cluster_experiments/experiment_analysis.py
```diff
@@ -301,6 +302,112 @@ def from_config(cls, config):
...
+            [x in self.cluster_cols for x in self.strata_cols]
+        ), f"strata should be a subset or equal to cluster_cols ({self.cluster_cols = }, {self.strata_cols = })"
+
+        df_pivot = df_grouped.pivot_table(
+            columns=self.treatment_col,
+            index=self.strata_cols,
+            values=self.target_col,
+        )
+
+        if df_pivot.isna().sum().sum() == 0:
```
david26694, line 375: what's this case doing?

tests/analysis/test_analysis.py
```diff
@@ -56,6 +57,38 @@ def test_ttest(analysis_df_diff):
...
+
+def test_paired_ttest_preprocessing():
+
+    analyser = PairedTTestClusteredAnalysis(
+        cluster_cols=["country_code", "city_code"], strata_cols=["country_code"]
+    )
+
+    df_pivot = analyser.preprocessing(df=generate_clustered_data())
+
+    assert df_pivot.isna().sum().sum() == 0, "Unexpected nas in pivot"
```
david26694, line 89: do you think we can make more extensive cheks in here? like, we have a pretty specific image of what should df_pivot be right? some things like:
* number of rows
* average target per country in A and B (we should change the target in the example df for it not to be trivial)

I think it is worth be super sure of what this is doing by more unit testing

david26694 (changes requested):
Please also add the method in the readme, and in the `__init__`

cluster_experiments/experiment_analysis.py
Gabrielcidral1: I ran pytest (with df['cluster']) and it's passing. 

image

Will fix it anyway

cluster_experiments/experiment_analysis.py
Gabrielcidral1: removing for example cluster 63 from analysis

image

Gabrielcidral1:
Hey David, I think we have a problem because the strata on splitter can be different than the strata on analysis. For example, If my clusters are ['Date', 'Country'], I might add a strata with 'day of week'. However, when doing the analysis, I might only want to have strata = ['Country']. In the power config at the moment only accepts one strata that will apply to all classes.

cluster_experiments/experiment_analysis.py
david26694: alright, could you comment this somehow? that his is to handle cases where the pairing doesn't exist

david26694:
> Hey @david26694 , I think we have a problem because the strata on splitter can be different than the strata on analysis. For example, If my clusters are ['Date', 'Country'], I might add a strata with 'day of week'. However, when doing the analysis, I might only want to have strata = ['Country']. In the power config at the moment only accepts one strata that will apply to all classes.

Why would you do this? There are 3 stratification options in the analysis given that cluster cols = [date, country] and strata_col = [country, day of week]. 
* strata col = day of week. Why would you do this? you have a calendar that generates day of week + country pairs, why would you lose part of this stratification by just pairing days of week?
* strata col = country. Why would you do this? you have a calendar that generates day of week + country pairs, why would you lose part of this stratification by just pairing days of week?
* strata col = [country, day of week]. This is the most reasonable option imo

cluster_experiments/experiment_analysis.py
```diff
@@ -301,6 +302,114 @@ def from_config(cls, config):
...
+        """Returns the p-value of the analysis
+        Arguments:
+            df: dataframe containing the data to analyze
+            verbose (Optional): bool, prints the regression summary if True
+        """
+
+        df_pivot = self._preprocessing(df=df)
+
+        if verbose:
+            print(f"performing paired t test in this data \n {df_pivot} \n")
```
david26694, line 392: can you put this print below? so that we only have 1 if verbose, and decouple printing and calculation (I know it's a bit weird because you're lying). if you don't like it, I suggest to move this part of the code to _preprocessing

cluster_experiments/experiment_analysis.py
```diff
@@ -301,6 +302,114 @@ def from_config(cls, config):
...
+    df = pd.DataFrame({
+        'x': [1, 2, 3, 4, 0, 0, 1, 1],
+        'treatment': ["A", "B", "A", "B"] * 2,
+        'cluster': [1, 2, 3, 4, 1, 2, 3, 4],
+    })
+
+    PairedTTestClusteredAnalysis(
+        cluster_cols='cluster',
+        strata_cols='cluster',
+        target_col=df['x'],
```
david26694, line 331: ah I know why, you have to add your class in tests/test_docs, and this test will fail. Could you also add in in the readme, in the `__init__`, and increase the library version in setup.py to 0.5.1?

docs/paired_ttest.ipynb
```diff
@@ -0,0 +1,234 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "source": [
+    "This notebook shows how to use the switchback module. In particular, it shows how to create a PowerAnalysis object with a switchback splitter, using a time column and 30 min splits.\n",
```
david26694, line 6: can you remove this? I don't think it applies. You also need to run the notebook if you want to see the results of the notebook in the docs

david26694: you also need to add the notebook path in mkdocs.yml if you'd like to see it in the docs. You can type `make docs-serve` if you want to test the docs output locally

david26694 (changes requested):
Last request changes :)

david26694 (changes requested):
Last changes:
* Add paired ttest it in readme
* Remove test_verbose

david26694 (approved):
gj!

PR #91, opened 2023-03-09
Two-sided washover

david26694:


pablobd:
@david26694 do you still plan to add this feature? I think it's a good one

david26694:
> @david26694 do you still plan to add this feature? I think it's a good one

yes that's my idea, couldn't find the time though

PR #102, opened 2023-04-28
Add MLM analysis

alexanderhensel:
Add Mixed Linear Model experiment analysis

Note: It could be a good idea to have the `fit` method in the interface also for other statsmodels analyses. This could also allow for avoiding `fit` to be called multiple times when separately executing `analysis_point_estimate` and `get_pvalue`. But I decided not to touch the interface in this PR.

### DIFF
--- a/cluster_experiments/experiment_analysis.py
+++ b/cluster_experiments/experiment_analysis.py
@@ -634,3 +634,13 @@ class MLMExperimentAnalysis(ExperimentAnalysis):
         results_mlm = self.fit_mlm(df)
         # negative sign, as we use the treatment as reference (instead of the baseline)
         return -results_mlm.params[self.mlm_treatment_label]
+
+    def get_pvalue(self, df: pd.DataFrame) -> float:
+        """Returns the p-value of the analysis
+
+        Arguments:
+            df: dataframe containing the data to analyze
+        """
+        df = df.copy()
+        self._data_checks(df=df)
+        return self.analysis_pvalue(df)
--- a/cluster_experiments/experiment_analysis.py
+++ b/cluster_experiments/experiment_analysis.py
@@ -586,23 +586,14 @@ class MLMExperimentAnalysis(ExperimentAnalysis):
             treatment=treatment,
             covariates=covariates,
         )
-        # the specification with C() ensures consistent ordering of treatment and non-treatment in the model
-        self.mlm_treatment_spec = (
-            f"C({self.treatment_col}, Treatment(reference='{self.treatment}'))"
-        )
-        self.mlm_treatment_label = None
-        self.formula = f"{self.target_col} ~ {self.mlm_treatment_spec}"
-        if self.covariates:
-            self.formula = f"{self.formula} + {' + '.join(self.covariates)}"
+        self.regressors = [self.treatment_col] + self.covariates
+        self.formula = f"{self.target_col} ~ {' + '.join(self.regressors)}"
 
         self.re_formula = None
         self.vc_formula = None
 
     def fit_mlm(self, df: pd.DataFrame) -> sm.MixedLM:
         """Returns the fitted MLM model"""
-        self.mlm_treatment_label = (
-            f"{self.mlm_treatment_spec}[T.{self._get_non_treatment_label(df)}]"
-        )
         return sm.MixedLM.from_formula(
             formula=self.formula,
             data=df,
@@ -610,9 +601,6 @@ class MLMExperimentAnalysis(ExperimentAnalysis):
             re_formula=self.re_formula,
             vc_formula=self.vc_formula,
         ).fit()
-
-    def _get_non_treatment_label(self, df):
-        return (set(df[self.treatment_col]) - set([self.treatment])).pop()
 
     def analysis_pvalue(self, df: pd.DataFrame, verbose: bool = False) -> float:
         """Returns the p-value of the analysis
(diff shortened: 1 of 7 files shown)


pablobd (commented):
It would be good to update the docs too and the Readme

cluster_experiments/experiment_analysis.py
```diff
@@ -539,3 +539,108 @@ def from_config(cls, config):
...
+
+    def analysis_point_estimate(self, df: pd.DataFrame, verbose: bool = False) -> float:
+        """Returns the point estimate of the analysis
+        Arguments:
+            df: dataframe containing the data to analyze
+            verbose (Optional): bool, prints the regression summary if True
+        """
+        results_mlm = self.fit_mlm(df)
+        # negative sign, as we use the treatment as reference (instead of the baseline)
+        return -results_mlm.params[self.mlm_treatment_label]
```
david26694, line 636: for consistency with the other classes in the library, could we return the usual
```python
        return results_mlm.params[self.mlm_treatment_label]
```
and you change the sign in your code?

cluster_experiments/experiment_analysis.py
```diff
@@ -539,3 +539,108 @@ def from_config(cls, config):
...
+
+    def get_pvalue(self, df: pd.DataFrame) -> float:
```
david26694, line 638: why do you need to override this?

david26694 (changes requested):
Can you add your class to:
* tests/test_docs
* __init__
?

Can you increase the library version in setup.py?

cluster_experiments/experiment_analysis.py
alexanderhensel: I could also have overridden `_get_binary_treatment` to just return the input dataframe, but I felt that having a method called `_get_binary_treatment` that's just the identity function is a bit misleading and may cause problems elsewhere... Would you prefer this option?

cluster_experiments/experiment_analysis.py
alexanderhensel: Hmmm my thinking was that applying the sign here is precisely so that the output of the method is aligned with the output of the same method in the other classes... I feel that it isn't optimal for the user to always have to think about reversing the sign when using this class. What do you think? Do you have another argument to convince me here?^^

alexanderhensel:
Thanks @david26694 for the comments! Which `__init__` are you referring to? Cause I think it's already in the main `__init__` and the `__init__`'s for the separate modules are all empty, so I didn't want to change them.

cluster_experiments/experiment_analysis.py
david26694: what's the issue of not overriding any of the 2?

cluster_experiments/experiment_analysis.py
david26694: so, does `-results_mlm.params[self.mlm_treatment_label]` give an estimate of the ATE? if I run, with the same data a ClusteredOls and a MixedLM, do I get similar results of get_point_estimate? or is the sign changed?

david26694:
> Thanks @david26694 for the comments! Which `__init__` are you referring to? Cause I think it's already in the main `__init__` and the `__init__`'s for the separate modules are all empty, so I didn't want to change them.

it's good now, I may have not seen it

cluster_experiments/experiment_analysis.py
alexanderhensel: Correct, with the negative sign the point estimate result is the same as for ClusteredOLS. When I tried ClusteredOLS by the way, I found that using the df that's in the docstring it doesn't work: The reason is that in the params of the fitted ols model, the code is looking for the label `"treatment"`, but actually it's `"treatment[T.B]"`: This is precisely the case I wanted to avoid in my code and where the negative sign results from (see the comment I left [here](https://github.com/david26694/cluster-experiments/blob/7eb983ef08de088caf9ba25c3a71110582035946/cluster_experiments/experiment_analysis.py#L589)). I'm not exactly sure how `sm.OLS` behaves, but in the case of `sm.MixedLM`, the assignment of which of the labels in the `"treatment"` column is considered to be the "baseline" and which is the "treated" is (I believe) simply based on which of the two labels ("A" or "B" in the case of the test df) appears first. So I explicitly made it deterministic to always be the "treated" ("B" in our case) first (the only reason being that we already have this label in `self.treatment`, whereas we don't have any information on the "baseline label"). In order to then convert to ATE, I have to take the negative sign.
Perhaps it would be good to do something similar in the `ClusteredOLSAnalysis` class?

alexanderhensel:
@david26694 I see now how I can change things, given that the interface's `analysis_pvalue` and `analysis_point_estimate` require 0-1 labels. But tbh I got slightly confused by this and the method naming here, but I'm okay changing my code to make it more rigid to fit exactly the docstring of the interface if this is your preference.

david26694:
> @david26694 I see now how I can change things, given that the interface's `analysis_pvalue` and `analysis_point_estimate` require 0-1 labels. But tbh I got slightly confused by this and the method naming here, but I'm okay changing my code to make it more rigid to fit exactly the docstring of the interface if this is your preference.

Yeah, agree that the method naming is not great. If you think you can make it fit the docstring of the interface, I would rather have it like that; and in another PR we can discuss about the interfaces

david26694 (approved):
lgtm

PR #107, opened 2023-05-08
feat: add seed to reproduce simulations

pablobd:
In this PR, I add the possibility to set a seed, so that simulations are reproducible. Also it will allow to make more fair comparisons betweens methods of analysis.

The main sources of stochasticity that are controlled with the seed are:

1. The splitter
2. The binary perturbator

### DIFF
--- a/cluster_experiments/power_analysis.py
+++ b/cluster_experiments/power_analysis.py
@@ -1,6 +1,8 @@
 import logging
+import random
 from typing import Dict, Generator, Iterable, List, Optional, Tuple
 
+import numpy as np
 import pandas as pd
 from sklearn.base import BaseEstimator
 from tqdm import tqdm
@@ -41,6 +43,7 @@ class PowerAnalysis:
         n_simulations: Number of simulations to run.
         alpha: Significance level.
         features_cupac_model: Covariates to be used in cupac model
+        seed: Optional. Seed to use for the splitter.
 
     Usage:
     ```python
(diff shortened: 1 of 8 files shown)


cluster_experiments/power_analysis.py
```diff
@@ -115,6 +119,10 @@ def __init__(
             target_col=target_col,
             features_cupac_model=features_cupac_model,
         )
+        if seed:
```
david26694, line 122:
```suggestion
        if seed is not None:
```

perhaps an edge case, but if seed = 0, won't seed anything

tests/power_analysis/test_seed.py
```diff
@@ -0,0 +1,37 @@
+import numpy as np
+
+from cluster_experiments.perturbator import Perturbator
+from cluster_experiments.power_analysis import PowerAnalysis
+
+
+def get_config(perturbator: Perturbator) -> dict:
```
david26694, line 7:
```suggestion
def get_config(perturbator: str) -> dict:
```

I think it's a string

david26694 (commented):
couple of minor comments

cluster_experiments/power_analysis.py
pablobd: Good catch! I changed it to your suggestion

tests/power_analysis/test_seed.py
pablobd: True, changed it too

david26694 (approved):
lgtm!

alexanderhensel (approved):
LGTM

PR #110, opened 2023-05-09
Pb/add relative stochastic perturbators

pablobd:
In this PR, I am adding:
- `RelativePositivePerturbator`: a refactored version of a multiplicative perturbator originally developed by @DaniGate in a dev branch
- `StochasticPerturbator`: additive stochastic perturbator, it perturbs by sampling from a Normal distribution
- `StochasticRelativePositivePerturbator`: a fancy perturbator, multiplicative and stochastic, sampling from a Beta distribution between (ranged between 0 and 1). The beta is conveniently reparametrized with the mean and scale.

100% of diff lines are tested having an effect of increasing coverage above 98%.

image

### DIFF
--- a/cluster_experiments/perturbator.py
+++ b/cluster_experiments/perturbator.py
@@ -276,11 +276,11 @@ class StochasticRelativePositivePerturbator(
         """
         Usage:
         ```python
-        from cluster_experiments.perturbator import RelativePositivePerturbator
+        from cluster_experiments.perturbator import StochasticRelativePositivePerturbator
         import pandas as pd
         df = pd.DataFrame({"target": [1, 2, 3], "treatment": ["A", "B", "A"]})
         perturbator = StochasticRelativePositivePerturbator()
-        # Increase target metric by 50%
+        # Increase target metric by 20%
         perturbator.perturbate(df, average_effect=0.2)
         # returns pd.DataFrame({"target": [1, 3, 3], "treatment": ["A", "B", "A"]})
         ```
--- a/cluster_experiments/perturbator.py
+++ b/cluster_experiments/perturbator.py
@@ -1,6 +1,7 @@
 from abc import ABC, abstractmethod
 from typing import Optional
 
+import numpy as np
 import pandas as pd
 
 
@@ -17,7 +18,7 @@ class Perturbator(ABC):
 
     Arguments:
         average_effect: The average effect of the treatment
-        treatment: name of the treatment to use as the treated group
+        target_col: name of the target_col to use as the treated group
         treatment_col: The name of the column that contains the treatment
         treatment: name of the treatment to use as the treated group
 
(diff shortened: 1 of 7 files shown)


cluster_experiments/perturbator.py
```diff
@@ -90,6 +91,206 @@ def perturbate(
...
+
+    Arguments:
+        average_effect (Optional[float], optional): the average effect of the treatment. Defaults to None.
+        target_col (str, optional): name of the target_col to use as the outcome. Defaults to "target".
+        treatment_col (str, optional): the name of the column that contains the treatment. Defaults to "treatment".
+        treatment (str, optional): name of the treatment to use as the treated group. Defaults to "B".
+        scale (Optional[float], optional): the scale of the effect distribution. Defaults to None.
+    """
+
+    def __init__(
```
david26694, line 108: I think you need to add a from_config method and pass the scale in there, if not it won't be sent when building from config

david26694 (changes requested):
Can you add the new classes to the tests/test_docs?

cluster_experiments/perturbator.py
pablobd: nice catch! it's fixed now

alexanderhensel (approved):
LGTM. I'm wondering if it would be worth adding a parameter so that every perturbator could be used both as "additive" and as "relative", without the need to define two separate classes.

pablobd:
> 



> LGTM. I'm wondering if it would be worth adding a parameter so that every perturbator could be used both as "additive" and as "relative", without the need to define two separate classes.

It's better to keep classes small and focused on what they should do

cluster_experiments/perturbator.py
```diff
@@ -90,6 +91,217 @@ def perturbate(
...
+        import pandas as pd
+        df = pd.DataFrame({"target": [1, 2, 3], "treatment": ["A", "B", "A"]})
+        perturbator = StochasticRelativePositivePerturbator()
+        # Increase target metric by 20%
+        perturbator.perturbate(df, average_effect=0.2)
+        # returns pd.DataFrame({"target": [1, 3, 3], "treatment": ["A", "B", "A"]})
+        ```
+        """
+        df = df.copy().reset_index(drop=True)
+        average_effect = self.get_average_effect(average_effect)
```
david26694, line 289: Perhaps out of the scope of this PR, but since you're working on the perturbators, wdyt about renaming get_average_effect to _get_average_effect?

david26694: (for all perturbators)

david26694:
Some minor changes

cluster_experiments/perturbator.py
pablobd: Thanks for the comment, since it's invoked by other classes, it can't be a private method. Therefore, it's better like this.

cluster_experiments/perturbator.py
```diff
@@ -90,6 +91,217 @@ def perturbate(
...
+        df = pd.DataFrame({"target": [1, 2, 3], "treatment": ["A", "B", "A"]})
+        perturbator = RelativePositivePerturbator()
+        # Increase target metric by 50%
+        perturbator.perturbate(df, average_effect=0.5)
+        # returns pd.DataFrame({"target": [1, 3, 3], "treatment": ["A", "B", "A"]})
+        ```
+        """
+        df = df.copy().reset_index(drop=True)
+        average_effect = self.get_average_effect(average_effect)
+        self._assert_multiplicative_effect(df, average_effect)
```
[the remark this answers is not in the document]
pablobd: done

cluster_experiments/perturbator.py
```diff
@@ -90,6 +91,217 @@ def perturbate(
...
+            raise ValueError(f"scale must be positive, got {scale}")
+        return scale
+
+    def _get_number_of_treated(self, df: pd.DataFrame) -> int:
+        """Get the number of treated instances in the dataframe"""
+        return (df[self.treatment_col] == self.treatment).sum()
+
+    def _sample_normal_effect(
+        self, average_effect: float, scale: float, n: int
+    ) -> float:
```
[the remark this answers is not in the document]
pablobd: good catch! done

david26694 (approved):
amazing!

PR #120, opened 2023-05-16
feat: allow washover-time int

pablobd:
In this PR, I am adding warnings when superfluous params are provided to the `PowerConfig`. Superfluous params are those that have no effect in the power simulations run. It also overrides these superfluous params to the default values. In this way, if config1 and config2 differ in superfluous params, this code `config1 == config2` will eval to True.

### DIFF
--- a/cluster_experiments/power_config.py
+++ b/cluster_experiments/power_config.py
@@ -169,7 +169,7 @@ class PowerConfig:
                     "Overriding treatments to None."
                 )
 
-        if "stratified" not in self.splitter:
+        if "stratified" not in self.splitter and "ttest" not in self.analysis:
             if self._are_different(self.strata_cols, None):
                 self.strata_cols = None
                 logging.warning(
@@ -177,25 +177,25 @@ class PowerConfig:
                     "Overriding strata_cols to None."
                 )
 
-        if "cupac" not in self.analysis:
-            if self._are_different(self.agg_col, ""):
-                self.agg_col = ""
-                logging.warning(
-                    f"{self.agg_col = } has no effect with {self.analysis = }."
-                    "Overriding agg_col to None."
-                )
-            if self._are_different(self.smoothing_factor, 20):
-                self.smoothing_factor = 20
-                logging.warning(
-                    f"{self.smoothing_factor = } has no effect with {self.analysis = }."
-                    "Overriding smoothing_factor to 20."
-                )
-            if self._are_different(self.features_cupac_model, None):
-                self.features_cupac_model = None
-                logging.warning(
-                    f"{self.features_cupac_model = } has no effect with "
-                    f"{self.analysis = }. Overriding features_cupac_model to None."
-                )
+        # if "cupac" not in self.analysis:
+        #     if self._are_different(self.agg_col, ""):
+        #         self.agg_col = ""
+        #         logging.warning(
+        #             f"{self.agg_col = } has no effect with {self.analysis = }."
+        #             "Overriding agg_col to None."
+        #         )
+        #     if self._are_different(self.smoothing_factor, 20):
+        #         self.smoothing_factor = 20
+        #         logging.warning(
+        #             f"{self.smoothing_factor = } has no effect with {self.analysis = }."
+        #             "Overriding smoothing_factor to 20."
+        #         )
+        #     if self._are_different(self.features_cupac_model, None):
+        #         self.features_cupac_model = None
+        #         logging.warning(
+        #             f"{self.features_cupac_model = } has no effect with "
+        #             f"{self.analysis = }. Overriding features_cupac_model to None."
+        #         )
 
         if "ttest" in self.analysis:
             if self._are_different(self.covariates, None):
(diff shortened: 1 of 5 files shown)


cluster_experiments/power_config.py
```diff
@@ -29,7 +31,7 @@
 )
 
 
-@dataclass
+@dataclass(frozen=True)
```
david26694, line 34: what does this do?

david26694 (approved):
would you add a test for the int case? or is it too dummy?

cluster_experiments/power_config.py
pablobd: This makes the object immutable and hashable.... that means that we can have a set of configs or they can be the key of a dictionary. On the other hand, people can't change the arguments of the class once instantiated

pablobd:
> would you add a test for the int case? or is it too dummy?

Yes, I can. In which test file would you add it?

david26694:
> > would you add a test for the int case? or is it too dummy?
> 
> Yes, I can. In which test file would you add it?

I guess in https://github.com/david26694/cluster-experiments/blob/main/tests/power_analysis/test_switchback_power.py
or https://github.com/david26694/cluster-experiments/blob/main/tests/splitter/test_washover.py, but perhaps a new one  in splitter/ makes sense too

cluster_experiments/power_config.py
```diff
@@ -122,6 +126,88 @@ class PowerConfig:
 
     seed: Optional[int] = None
 
+    def __post_init__(self):
```
david26694, line 129: Wdyt about adding this method:

```python
    def set_and_log(self, attr, value, other_attr):
        setattr(self, attr, value)
        logging.warning(
            f"{attr} = {getattr(self, attr)} has no effect with "
            f"{other_attr} = {getattr(self, other_attr)}. "
            f"Overriding {attr} to {value}."
        )
```

then doing:

```python
        if "switchback" not in self.splitter:
            if self._are_different(self.switch_frequency, None):
                self.set_and_log("switch_frequency", None, "splitter")
```

cluster_experiments/power_config.py
```diff
@@ -122,6 +126,88 @@ class PowerConfig:
...
+            if self._are_different(self.scale, None):
+                self.scale = None
+                logging.warning(
+                    f"{self.scale = } has no effect with {self.perturbator = }."
+                    "Overriding scale to None."
+                )
+
+        if "balanced" not in self.splitter:
+            if self._are_different(self.treatments, None):
+                self.treatments = None
```
david26694, line 166: I think treatments should work with any splitter, not only balanced splitters

cluster_experiments/power_config.py
```diff
@@ -122,6 +126,88 @@ class PowerConfig:
...
+                logging.warning(
+                    f"{self.treatments = } has no effect with {self.splitter = }."
+                    "Overriding treatments to None."
+                )
+
+        if "stratified" not in self.splitter and "ttest" not in self.analysis:
```
david26694, line 172: We can even be more precise and do:
```suggestion
        if "stratified" not in self.splitter and "paired_ttest" not in self.analysis:
```

cluster_experiments/power_config.py
```diff
@@ -122,6 +126,88 @@ class PowerConfig:
...
+            if self._are_different(self.strata_cols, None):
+                self.strata_cols = None
+                logging.warning(
+                    f"{self.strata_cols = } has no effect with {self.splitter = }."
+                    "Overriding strata_cols to None."
+                )
+
+        # if "cupac" not in self.analysis:
```
david26694, line 180: This chould be:
```python
if cupac_model == ""
```

david26694 (changes requested):
Some minor changes, it also needs to increase version in setup. At some point I'd add the tests about the `_post_init_`, maybe we can do in a future PR

cluster_experiments/power_config.py
pablobd: So, removing this block entirely!

cluster_experiments/power_config.py
pablobd: thanks, I changed it

cluster_experiments/power_config.py
pablobd: thanks. It's done now

cluster_experiments/power_config.py
pablobd: Looks nice! I'll give it a try

cluster_experiments/power_config.py
```diff
@@ -122,6 +126,48 @@ class PowerConfig:
 
     seed: Optional[int] = None
 
+    def __post_init__(self):
+        if "switchback" not in self.splitter:
+            if self._are_different(self.switch_frequency, None):
+                self._set_and_log("swithc_frequency", None, "splitter")
```
david26694, line 132:
```suggestion
                self._set_and_log("switch_frequency", None, "splitter")
```

david26694 (changes requested):
one last change, can you increase setup version?

cluster_experiments/power_config.py
pablobd: I got this and other minor errors thanks to the new tests added  🚀

david26694 (approved):
lgtm! let's see if it's released :)

PR #121, opened 2023-05-17
fix: type hint config

pablobd:
Docs where failing to build as a future version of python was needed.


david26694 (approved):
I'd also increase the version in the setup, minimal change but to avoid issues

PR #127, opened 2023-05-23
Enable power analysis for backtests

alexanderhensel:
Add `BacktestSplitter`, `BetaRelativePerturbator` and `SegmentedBetaRelativePerturbator`.


cluster_experiments/perturbator.py
```diff
@@ -350,6 +351,274 @@ def _sample_beta_effect(
...
+        return df
+
+    @staticmethod
+    def _check_scale_range(range_min: float, range_max: float):
+        if range_min < -1:
+            raise ValueError(f"range_min needs to be greater than -1, got {range_min}")
+        if range_min >= range_max:
+            raise ValueError(
+                f"range_min needs to be smaller than range_max, got "
+                f"range_min={range_min} and range_max={range_max}"
```
david26694, line 445: more compact:
```suggestion
                f"{range_min = } and {range_max = }"
```

cluster_experiments/perturbator.py
```diff
@@ -350,6 +351,274 @@ def _sample_beta_effect(
...
+            range_max=config.range_max,
+        )
+
+
+class ClusteredBetaRelativePerturbator(BetaRelativePositivePerturbator):
+    """
+    A stochastic Perturbator for continuous targets that applies a sampled
+    effect from the Beta distribution. It applies the effect multiplicatively
+    and based on given clusters.
+    For each cluster, the average cluster effect is sampled from a beta
```
david26694, line 509: is there another name which is not cluster? I think we may have a lot of confusion with clusters from the cluster-randomized trial, in splitter and analysis

cluster_experiments/perturbator.py
```diff
@@ -350,6 +351,274 @@ def _sample_beta_effect(
...
+
+        self.cluster_col = "_cluster_" + "_".join(cluster_cols)
+        if self.cluster_col in df.columns:
+            raise ValueError(
+                f"Cannot use self.cluster_col={self.cluster_col} as clustering column, as it already exists in the input dataframe!"
+            )
+
+        df = df.copy()
+        cluster = pd.Series("", index=df.index)
+        for col in cluster_cols:
```
david26694, line 581: can't this be done with a sum? like this: https://github.com/david26694/cluster-experiments/blob/main/cluster_experiments/experiment_analysis.py#L46

cluster_experiments/random_splitter.py
```diff
@@ -481,3 +481,56 @@ def from_config(cls, config) -> "StratifiedSwitchbackSplitter":
             splitter_weights=config.splitter_weights,
             washover=washover_cls.from_config(config),
         )
+
+
+class BacktestSplitter(RandomSplitter):
+    """
+    Doesn't actually split the data, but duplicates all rows for all treatments.
```
david26694, line 488: in this case, wdyt about renaming to something more specific about the sampler? like RepeatedSampler or something like that

cluster_experiments/random_splitter.py
```diff
@@ -481,3 +481,56 @@ def from_config(cls, config) -> "StratifiedSwitchbackSplitter":
...
+
+    def assign_treatment_df(
+        self,
+        df: pd.DataFrame,
+    ) -> pd.DataFrame:
+        df = df.copy()
+
+        dfs = []
+        for treatment in self.treatments:
+            df_treat = df.copy()
```
david26694, line 524:
```suggestion
            df_treat = df.copy().assign(**{self.treatment_col: treatment})
```

cluster_experiments/perturbator.py
alexanderhensel: Thanks! I wasn't aware of this syntax!

cluster_experiments/random_splitter.py
alexanderhensel: done

cluster_experiments/perturbator.py
```diff
@@ -350,6 +351,274 @@ def _sample_beta_effect(
...
+    effect_transformed ~ beta(a/abs(mu), b/abs(mu))
+    effect = effect_transformed * (range_max - range_min) + range_min
+    ```
+
+    Arguments:
+        average_effect (Optional[float], optional): the average effect of the treatment. Defaults to None.
+        target_col (str, optional): name of the target_col to use as the outcome. Defaults to "target".
+        treatment_col (str, optional): the name of the column that contains the treatment. Defaults to "treatment".
+        treatment (str, optional): name of the treatment to use as the treated group. Defaults to "B".
+        scale (Optional[float], optional): the scale of the effect distribution. Defaults to None.
```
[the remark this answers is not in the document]
alexanderhensel: improved docstring

cluster_experiments/perturbator.py
alexanderhensel: So much nicer! Thanks!

cluster_experiments/perturbator.py
alexanderhensel: Do you have a suggestion? I agree that the clustering should be distinguished better on the 3 levels of splitter, perturbator and analysis (e.g. as I've tried to start doing in the power config). But I'm not sure how helpful a new name can be, considering that the purpose is really the same (just on a different level). But I may also just be thinking that because I can't see another option that I like^^

cluster_experiments/perturbator.py
```diff
@@ -350,6 +351,273 @@ def _sample_beta_effect(
...
+            )
+
+        return df.copy().assign(
+            **{self.cluster_col: df[cluster_cols].astype(str).sum(axis=1)}
+        )
+
+    def perturbate(
+        self,
+        df: pd.DataFrame,
+        cluster_cols: Optional[List[str]] = None,
```
david26694, line 587: I'd rather have the perturbate method keep the signature of the abstract method. In this case, it would be sending only df and average_effect, and cluster_cols can only be defined in the init. Wdyt? do you see much value in adding cluster_cols in here?

david26694: (in this case, I'd move some of the code of the _set_cluster_col method to the init)

cluster_experiments/perturbator.py
david26694: The purpose is really the same -> this is what I don't fully get. Both in the splitter and analysis, cluster_cols have a very precise meaning: it is the unit of treatment assignment: if in your experiment design, you assign cities to the treatment, then the clusters should be cities. If you assign stores, it should be stores.

In this case, I don't see how the clusters are related to units of treatment assignment. For instance, I understand you could be applying the treatment to cities and have the perturbation work differently for different stores. 

Do you agree? if so, we can think of a better name

cluster_experiments/perturbator.py
```diff
@@ -350,6 +351,273 @@ def _sample_beta_effect(
...
+        average_effect: Optional[float] = None,
+    ) -> pd.DataFrame:
+        df = df.copy().reset_index(drop=True)
+        df = self._set_cluster_col(df, cluster_cols)
+
+        clusters = df[self.cluster_col].unique()
+        cluster_dfs = []
+        for cluster in clusters:
+            df_cluster = df[df[self.cluster_col] == cluster].copy()
```
david26694, line 596: perhaps you can add all the filter logic and perturbation in a single method

tests/perturbator/test_perturbator.py
```diff
@@ -201,6 +203,93 @@ def test_stochastic_relative_perturbate_scale_provided_is_used(average_effect, s
...
+def test_clustered_beta_relative_perturbate(average_effect):
+    range_min = -0.8
+    range_max = 5
+
+    df_clustered = generate_clustered_data()
+    pert = ClusteredBetaRelativePerturbator(
+        range_min=range_min, range_max=range_max, cluster_cols=["city_code"]
+    )
+    df_pert = pert.perturbate(
+        pd.concat([df_clustered for _ in range(100)]), average_effect=average_effect
```
david26694, line 249: do you have some kind of test that, when applying a certain average effect (let's say +20%), you have data in treatment with the target (on average) higher than the data in control by a 20%? if not, wdyt about adding it?

cluster_experiments/power_analysis.py
```diff
@@ -284,6 +284,8 @@ def _non_parallel_loop(
         n_detected_mde = 0
         for _ in tqdm(range(n_simulations), disable=not verbose):
             p_value = self._run_simulation((df, average_effect))
+            if verbose:
+                print(f"p_value of simulation run: {p_value}")
```
david26694, line 288: perhaps we can format this:
```suggestion
                print(f"p_value of simulation run: {p_value:.3f}")
```

david26694 (commented):
Some minor comments

cluster_experiments/perturbator.py
alexanderhensel: Agreed and changed!

cluster_experiments/power_analysis.py
alexanderhensel: done

cluster_experiments/perturbator.py
```diff
@@ -234,7 +235,7 @@ def check_target_is_not_constant_zero(
             (df[self.treatment_col] != self.treatment) | (df[self.target_col] == 0)
         ).mean()
         if 1.0 == treatment_zeros:
-            raise ValueError(
+            warnings.warn(
```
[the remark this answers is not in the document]
alexanderhensel: If you or @pablobd prefers me to revert this change, I'm okay to do so: In the end I simply avoided using this method in the `ClusteredBetaRelativePerturbator`

tests/perturbator/test_perturbator.py
alexanderhensel: Not sure I understand, cause I think that's exactly what this test does^^
The only catch is that I'm allowing for a given `average_effect`, I'm allowing the _observed_ average effect in the treatment to vary by +-50% from the given `average_effect`. I realize that this sounds like a lot, but given the small size of the test df and the groups in that df, the observed average effect can vary quite a bit (which is kind of the point).

cluster_experiments/perturbator.py
alexanderhensel: Good point, I agree. I guess I thought about it too much in terms of the backtests case, where most likely you'd like the clusters in both the perturbation and the analysis to be the same (clusters in the splitter don't matter in the backtest case, cause it's not splitting anything).
So, any suggestion for the naming?^^

tests/perturbator/test_perturbator.py
david26694: alright, good. it was not clear to me that this was what we were testing, but great then :)

cluster_experiments/perturbator.py
david26694: wdyt about groups, perturbation_groups, GroupedBetaRelativePerturbator, etc?

cluster_experiments/perturbator.py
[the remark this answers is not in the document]
david26694: thoughts @pablobd ? I have no strong opinions

cluster_experiments/perturbator.py
```diff
@@ -350,6 +351,266 @@ def _sample_beta_effect(
...
+        super().__init__(average_effect, target_col, treatment_col, treatment, scale)
+        self._range_min = range_min
+        self._range_max = range_max
+        self._cluster_cols = cluster_cols
+        self.cluster_col = self._get_cluster_col_name(cluster_cols)
+
+    def get_cluster_perturbator(
+        self, average_effect: Optional[float] = None
+    ) -> Perturbator:
+        average_effect = self.get_average_effect(average_effect)
```
david26694, line 548: do you need to do all of this for each perturbator? or only once at the perturbate method?

alexanderhensel: Hmm yeah you're right that much of this only has to be done once. I now separated out a `get_cluster_perturbator_fixed_params` method which is only run once, and then the `get_cluster_perturbator` is run for each cluster. It may make the code a bit more wordy, but if ever we want to create another clustered perturbator, we only have to adjust few methods and it should be clear which ones to adjust. I hope this goes in the direction of what you meant here:)

tests/perturbator/test_perturbator.py
pablobd: looks like a lot variability. Shouldn't it converge to 0 if repeated many times? Maybe you could do this, iterate and check that in average it's smaller than epsilon. WDYT?

pablobd (commented):
I left a few comments. Let me know what you think.

cluster_experiments/perturbator.py
```diff
@@ -350,6 +351,266 @@ def _sample_beta_effect(
...
+    def _sample_scaled_beta_effect(
+        self, average_effect: float, scale: float, n: int
+    ) -> np.ndarray:
+        average_effect_inv_transf = self._inv_transform_to_range(average_effect)
+        scale_inv_transf = self._inv_transform_to_range(scale)
+        a = average_effect_inv_transf / (scale_inv_transf * scale_inv_transf)
+        b = (1 - average_effect_inv_transf) / (scale_inv_transf * scale_inv_transf)
+
+        # the division by abs(average_effect) makes the distribution more concentrated around the mean
+        beta = np.random.beta(a / abs(average_effect), b / abs(average_effect), n)
```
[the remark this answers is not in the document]
alexanderhensel: Yes I did check this. It's quite clear also from the formula for variance of the beta distribution, which is more or less proportional to `O(1 /(a + b))`, so for larger a and b the variance decreases. I now changed this slightly to only divide by `abs(average_effect)` if `average_effect < 1`.

cluster_experiments/perturbator.py
alexanderhensel: Changing to `Segmented`

cluster_experiments/power_config.py
```diff
@@ -100,6 +103,9 @@ class PowerConfig:
     # Perturbator
     average_effect: Optional[float] = None
     scale: Optional[float] = None
+    range_min: Optional[float] = None
+    range_max: Optional[float] = None
+    cluster_cols_perturbator: Optional[List[str]] = None
```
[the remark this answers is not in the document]
alexanderhensel: done

tests/perturbator/test_perturbator.py
alexanderhensel: @pablobd Good point, I changed it now to iterate a few times to decrease the variability. Not doing anything more fancy, like checking that the variability decreases with more runs, also because it slows down the unit tests. But let me know if you think it's important and I can do it.

david26694 (approved):
Great job!

cluster_experiments/perturbator.py
```diff
@@ -350,6 +351,266 @@ def _sample_beta_effect(
...
 
 
+class BetaRelativePerturbator(NormalPerturbator, RelativePositivePerturbator):
+    """
+    A stochastic Perturbator for continuous targets that applies a  sampled
+    effect from a scaled Beta distribution. It applies the effect multiplicatively.
+
+    The sampled effect is defined for values in the specified range
+    (range_min, range_max). It's recommended to set range_min<0 and range_max>0
+    in a "symmetric way" around 0, such that log(1 + range_min) = -log(1 + range_max).
```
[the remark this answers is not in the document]
alexanderhensel: Improved, hope it's clearer now:)

cluster_experiments/perturbator.py
```diff
@@ -350,6 +351,266 @@ def _sample_beta_effect(
...
+    perturbations that relatively increase the target.
+
+    The number of samples with 0 as target remains unchanged.
+
+    The stochastic effect is sampled from a beta distribution with parameters mean and
+    variance, which is scaled to the range (range_min, range_max).
+    If variance is not provided, the variance is abs(mean).
+
+    ```
+    target -> target * (1 + effect), where effect ~ Beta(a, b)
```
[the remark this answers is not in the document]
alexanderhensel: It's really just a linear transformation that's applied to the usual beta distribution with support in [0, 1], so I think there's really nothing magical going on here^^
Not sure if there's a small typo in your comment and you meant to say to add a Demo, in which case yes, this would be a good idea. I never looked at the other Demos though and my doubt right now is that considering that this perturbator could still change a bit once I try to integrate it into the experimentation-analysis workflow, there's a risk that the Demo might become outdated without us realizing. Or is "Desmos" some tool or package that I'm not aware of?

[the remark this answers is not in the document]
pablobd: I meant Desmos actually: https://www.desmos.com/calculator/frnbtrv4bq

cluster_experiments/perturbator.py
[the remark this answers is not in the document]
pablobd: I understand you want to make it more concentrated, but it's not clear why... aren't you changing the variance that the user provided without telling it to the user?

PR #130, opened 2023-05-25
feat: raise error if strata cols are not provided

pablobd:


david26694 (changes requested):
increase version in setup and good to me

david26694 (approved):

PR #133, opened 2023-05-26
feat: add new mappings

david26694:


pablobd (approved):
LGTM as a first step to make naming more uniform - also no problem about invertability I think

PR #134, opened 2023-05-26
Add MissingArgumentException

alexanderhensel:
Maybe should be added for other parameters as well, to catch faulty configs at "init-time" rather than when trying to run the analysis (e.g. I just saw the PR, for which we could raise the error also in the config already: https://github.com/david26694/cluster-experiments/pull/130)

### DIFF
--- a/cluster_experiments/power_analysis.py
+++ b/cluster_experiments/power_analysis.py
@@ -284,8 +284,7 @@ class PowerAnalysis:
         n_detected_mde = 0
         for _ in tqdm(range(n_simulations), disable=not verbose):
             p_value = self._run_simulation((df, average_effect))
-            if verbose:
-                logging.info(f"p_value of simulation run: {p_value:.3f}")
+            logging.info(f"p_value of simulation run: {p_value:.3f}")
             n_detected_mde += p_value < alpha
 
         return n_detected_mde / n_simulations
--- a/cluster_experiments/power_analysis.py
+++ b/cluster_experiments/power_analysis.py
@@ -285,7 +285,7 @@ class PowerAnalysis:
         for _ in tqdm(range(n_simulations), disable=not verbose):
             p_value = self._run_simulation((df, average_effect))
             if verbose:
-                print(f"p_value of simulation run: {p_value:.3f}")
+                logging.info(f"p_value of simulation run: {p_value:.3f}")
             n_detected_mde += p_value < alpha
 
         return n_detected_mde / n_simulations
--- a/cluster_experiments/power_analysis.py
+++ b/cluster_experiments/power_analysis.py
@@ -284,7 +284,8 @@ class PowerAnalysis:
         n_detected_mde = 0
         for _ in tqdm(range(n_simulations), disable=not verbose):
             p_value = self._run_simulation((df, average_effect))
-            logging.info(f"p_value of simulation run: {p_value:.3f}")
+            if verbose:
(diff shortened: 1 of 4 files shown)


cluster_experiments/power_config.py
```diff
@@ -185,6 +189,9 @@ def __post_init__(self):
             if self._are_different(self.covariates, None):
                 self._set_and_log("covariates", None, "analysis")
 
+        if "segmented" in self.perturbator:
```
david26694, line 192: do you think it makes sense to raise this in the perturbator init? then we would raise when this is initialized from config or using the regular init

cluster_experiments/power_analysis.py
```diff
@@ -284,8 +284,7 @@ def _non_parallel_loop(
         n_detected_mde = 0
         for _ in tqdm(range(n_simulations), disable=not verbose):
             p_value = self._run_simulation((df, average_effect))
-            if verbose:
-                print(f"p_value of simulation run: {p_value:.3f}")
+            logging.info(f"p_value of simulation run: {p_value:.3f}")
```
david26694, line 287: I'd still add it instead the if verbose, wdyt?

david26694 (changes requested):

cluster_experiments/power_analysis.py
alexanderhensel: Hmm I'd say that using logging you already have a way of controlling the verbosity by setting the log level. In particular, I wanted to take it out from the `if verbose` since in the power_config we haven't parameterized the `verbose` setting and I wanted each simulation p_value to be logged when running simulations on argo (without having to add another parameter to the power_config).

cluster_experiments/power_config.py
alexanderhensel: I'm not sure: The difference is that in the specific perturbator class, the argument in question is already _not optional_, but it _is optional_ in the power_config if you choose not to use this specific perturbator.
So it feels like the error belongs more to the power_config, since in the estimator you can just rely on the signature. But we can also make our own convention here if you prefer, I don't have a suuuper strong opinion:)

david26694: you're right, fine in here

cluster_experiments/power_analysis.py
david26694: but you don't need to add another parameter right? you can call the power_analysis method with verbose=True and you get the printed logs, right?

david26694: (I agree that you have a way of controlling verbosity, but I think it's friendlier for users to pass the verbose=True or False)

alexanderhensel: Ah true, you're right, it doesn't have to be in the power_config itself!
About the `verbose` argument being "friendlier" than the log level, I don't know, seems a bit subjective^^ I don't have a strong opinion, but if we choose to use the `verbose` argument here then I'd maybe replace the usage of `logging` with a usual `print` statement, as otherwise there could be some confusion if the user sets `verbose=True` but the p_values don't show up. WDYT?

david26694: yes, agree!

david26694: then, we revert this, right?

alexanderhensel: okay, done

david26694 (approved):
gj!

PR #137, opened 2023-12-20
Allow users to run one tailed experiments 

Gabrielcidral1:
This PR aims to allow users to run 1 tailed power analyses. Currently, there's only the possibility of running 2 tailed. 

This is the [rationale for the adaptation](https://www.graphpad.com/support/faq/p-values-one-tail-or-two-tail-/):

> If the actual difference (effect) went in the predicted direction:
> The one-tail p-value is half the two-tail p-value. So if the two-tailed p-value is 0.1, the one-tailed p-value is 0.05.
> The two-tail p-value is twice the one-tail p-value.
> If the actual difference (effect) went opposite to the predicted direction:
> The one-tail p-value equals one minus half the two-tailed value. So if the two-tailed p-value is 0.1, the one-tailed p-value is 0.95
> The two-tail p-value is twice the one-tail p-value.

From my perspective, the best way is that in case of one-tailed experiments, the user inputs the desired direction (left or right, instead of just 'one-tailed'). In case we want accept 'one-sided', we could guess the side by the sign of the average_effect input. However, there would be a challenge as with current design the analysis class doesn't have the effect information.


Gabrielcidral1:
Why I can't see the tests? Do I need to deploy the branch?

cluster_experiments/experiment_analysis.py
```diff
@@ -111,6 +113,26 @@ def get_point_estimate(self, df: pd.DataFrame) -> float:
...
+        """Returns the p-value of the analysis
+        Arguments:
+            model_result: statsmodels result object
+            verbose (Optional): bool, prints the regression summary if True
+
+        """
+        treatment_effect = model_result.params[self.treatment_col]
+        p_value_half = model_result.pvalues[self.treatment_col] / 2
+
+        if self.hypothesis == "two_tailed" or self.hypothesis == "two-tailed":
```
david26694, line 126: is the or supposed to be in another way?
```if self.hypothesis == "two_tailed" or self.hypothesis == "two-tailed":```

File: cluster_experiments/experiment_analysis.py [hunk 2]
```diff
@@ -111,6 +113,26 @@ def get_point_estimate(self, df: pd.DataFrame) -> float:
         self._data_checks(df=df)
         return self.analysis_point_estimate(df)
 
+    def pvalue_based_on_hypothesis(self, model_result) -> float:
```
david26694, line 116: this only works for regression models, no? if it doesn't work for all kinds of analysis, I wouldn't put in here. I would use an external class or function that does this, and use composition instead of inheritance

cluster_experiments/experiment_analysis.py
```diff
@@ -33,12 +33,14 @@ def __init__(
...
         treatment: str = "B",
         covariates: Optional[List[str]] = None,
+        hypothesis: str = "two_tailed",
     ):
         self.target_col = target_col
         self.treatment = treatment
         self.treatment_col = treatment_col
         self.cluster_cols = cluster_cols
         self.covariates = covariates or []
+        self.hypothesis = hypothesis
```
david26694, line 43: as hypothesis can only have 3 values, I'd suggest using an enum somehow

cluster_experiments/experiment_analysis.py
david26694: ah now I see, maybe better in(two-tailed, two_tailed)

cluster_experiments/experiment_analysis.py
david26694: if we don't have a way to handle right-tailed or left-tailed for the t-test and paired t-test, we can always explicitely raise an error if left or right are passed to that test

tests/analysis/test_ols_analysis.py
```diff
@@ -14,5 +14,5 @@ def test_binary_treatment():
 
 def test_get_pvalue():
     analysis_df_full = pd.concat([analysis_df for _ in range(100)])
-    analyser = OLSAnalysis()
+    analyser = OLSAnalysis(hypothesis="left_tailed")
```
david26694, line 17: I'd add a separate test, and I'd test the functionality itself of the p-value transformer

david26694: as in, check that it is dividing by 2 or doing the other stuff when necessary

.pre-commit-config.yaml
```diff
@@ -12,7 +12,7 @@ repos:
     rev: 22.12.0
     hooks:
       - id: black
-        language_version: python3.8
+        language_version: python3.9
```
david26694, line 15: is this for your local developement? or is it because of github actions?

david26694 (changes requested):

david26694:
Another comment: to me it's a bit hard to remember what does left and right-sided test mean, I'd try to document very well which sign of the effect does it belong to, even in the readme

Gabrielcidral1:
> Another comment: to me it's a bit hard to remember what does left and right-sided test mean, I'd try to document very well which sign of the effect does it belong to, even creating some notebook in docs that shows the power of the 3 sides for a similar setting

I found a better way to do this. We can replicate scipy approach to it: 


>     alternative : {'two-sided', 'less', 'greater'}, optional
>         Defines the alternative hypothesis.
>         The following options are available (default is 'two-sided'):
> 
>         * 'two-sided': the means of the distributions underlying the samples
>           are unequal.
>         * 'less': the mean of the distribution underlying the first sample
>           is less than the mean of the distribution underlying the second
>           sample.
>         * 'greater': the mean of the distribution underlying the first
>           sample is greater than the mean of the distribution underlying
>           the second sample.
>

cluster_experiments/experiment_analysis.py
```diff
@@ -6,6 +6,7 @@
 import statsmodels.api as sm
 from pandas.api.types import is_numeric_dtype
 from scipy.stats import ttest_ind, ttest_rel
+from utils import HypothesisEntries
```
david26694, line 9: need to change this import

cluster_experiments/experiment_analysis.py
```diff
@@ -33,12 +34,14 @@ def __init__(
         treatment_col: str = "treatment",
         treatment: str = "B",
         covariates: Optional[List[str]] = None,
+        hypothesis: HypothesisEntries = HypothesisEntries.TWO_SIDED,
```
david26694, line 37: this needs to change to str, right?

cluster_experiments/experiment_analysis.py
```diff
@@ -111,6 +114,25 @@ def get_point_estimate(self, df: pd.DataFrame) -> float:
...
+        """
+        treatment_effect = model_result.params[self.treatment_col]
+        p_value_half = model_result.pvalues[self.treatment_col] / 2
+
+        if self.hypothesis == "less":
+            p_value = p_value_half if treatment_effect <= 0 else 1 - p_value_half
+        elif self.hypothesis == "greater":
+            p_value = p_value_half if treatment_effect >= 0 else 1 - p_value_half
+        elif self.hypothesis == "two-sided":
+            p_value = model_result.pvalues[self.treatment_col]
```
david26694, line 132: if we do this, I understand we are not using enum, right? Then I would remove the Enum code

david26694: also, an else clause is missing here raising an error

cluster_experiments/experiment_analysis.py
```diff
@@ -111,6 +114,25 @@ def get_point_estimate(self, df: pd.DataFrame) -> float:
...
+        Arguments:
+            model_result: statsmodels result object
+            verbose (Optional): bool, prints the regression summary if True
+
+        """
+        treatment_effect = model_result.params[self.treatment_col]
+        p_value_half = model_result.pvalues[self.treatment_col] / 2
+
+        if self.hypothesis == "less":
+            p_value = p_value_half if treatment_effect <= 0 else 1 - p_value_half
```
david26694, line 128: instead of computing p_value and returning in the end, I think you cna can just return here

cluster_experiments/experiment_analysis.py
```diff
@@ -111,6 +114,25 @@ def get_point_estimate(self, df: pd.DataFrame) -> float:
...
 
+    def pvalue_based_on_hypothesis(self, model_result) -> float:
+        """Returns the p-value of the analysis
+        Arguments:
+            model_result: statsmodels result object
+            verbose (Optional): bool, prints the regression summary if True
+
+        """
+        treatment_effect = model_result.params[self.treatment_col]
+        p_value_half = model_result.pvalues[self.treatment_col] / 2
```
david26694, line 125: perhaps it's easier to understand if you calculate p_value in here and then divide by 2 in the less and right cases

cluster_experiments/experiment_analysis.py
```diff
@@ -192,7 +216,8 @@ def analysis_pvalue(self, df: pd.DataFrame, verbose: bool = False) -> float:
         results_gee = self.fit_gee(df)
         if verbose:
             print(results_gee.summary())
-        return results_gee.pvalues[self.treatment_col]
+        p_value = self.pvalue_based_on_hypothesis(results_gee)
```
david26694, line 219:
```suggestion
        
        p_value = self.pvalue_based_on_hypothesis(results_gee)
```

david26694: (for consistency with others)

cluster_experiments/experiment_analysis.py
```diff
@@ -612,7 +647,9 @@ def analysis_pvalue(self, df: pd.DataFrame, verbose: bool = False) -> float:
         if verbose:
             print(results_mlm.summary())
 
-        return results_mlm.pvalues[self.treatment_col]
+        p_value = self.pvalue_based_on_hypothesis(results_mlm)
+
```
david26694, line 651:
```suggestion
```

for consistency

tests/analysis/test_ols_analysis.py
```diff
@@ -16,3 +17,10 @@ def test_get_pvalue():
     analysis_df_full = pd.concat([analysis_df for _ in range(100)])
     analyser = OLSAnalysis()
     assert analyser.get_pvalue(analysis_df_full) >= 0
+
+
+@pytest.mark.parametrize("hypothesis", ["one_sided", "two_sided"])
+def test_get_pvalue_hypothesis(hypothesis):
```
david26694, line 23: I would add tests for the pvalue_based_on_hypothesis method too, since it has some logic

david26694 (changes requested):

cluster_experiments/experiment_analysis.py
```diff
@@ -111,6 +115,25 @@ def get_point_estimate(self, df: pd.DataFrame) -> float:
...
+        treatment_effect = model_result.params[self.treatment_col]
+        p_value = model_result.pvalues[self.treatment_col]
+
+        if HypothesisEntries(self.hypothesis) == HypothesisEntries.LESS:
+            return p_value / 2 if treatment_effect <= 0 else 1 - p_value / 2
+        elif HypothesisEntries(self.hypothesis) == HypothesisEntries.GREATER:
+            return p_value / 2 if treatment_effect >= 0 else 1 - p_value / 2
+        elif HypothesisEntries(self.hypothesis) == HypothesisEntries.TWO_SIDED:
+            return p_value
+
```
david26694, line 136: I think we should add an explicit error if none of the conditions are met

cluster_experiments/experiment_analysis.py
```diff
@@ -33,12 +35,14 @@ def __init__(
         treatment_col: str = "treatment",
         treatment: str = "B",
         covariates: Optional[List[str]] = None,
+        hypothesis: str = "two-sided",
```
david26694, line 38: missing hypothesis in docstring

david26694: (from this class and some other classes)

tests/analysis/test_hypothesis.py
```diff
@@ -0,0 +1,56 @@
...
+    MLMExperimentAnalysis,
+    OLSAnalysis,
+    TTestClusteredAnalysis,
+)
+from tests.examples import analysis_df, generate_clustered_data
+
+
+@pytest.mark.parametrize("hypothesis", ["less", "greater", "two-sided"])
+@pytest.mark.parametrize("analysis_class", [OLSAnalysis])
+def test_get_pvalue_hypothesis(analysis_class, hypothesis):
```
david26694, line 16: I would also add some test where you initialize power from_dict and check that the hypothesis change is doing something

david26694 (changes requested):
Some things missing:
* There's some commented code in the notebook, I'd remove it
* You need to add the notebook in mkdocs.yml
* You need to increase the library version (I'd change from 0.10.2 to 0.11.0 or something like this)

david26694 (approved):

PR #147, opened 2024-02-01
143 add more effects in notebook

### LINKED ISSUE #143: Add more effects in notebook
https://david26694.github.io/cluster-experiments/analysis_with_different_hypothesis.html

This notebook needs more effects, otherwise this is not very representative:
imagen

Gabrielcidral1:
Adding more models and splitters to the hypothesis testing comparison


david26694 (commented):
Can you add even more effects?

Something like 10 points from 0 to 0.1, I think it'll make the visualization easier to understand
imagen

Gabrielcidral1:
I had a problem using plotnine version specified due to a conflict with numpy. I bumped plotnine locally to the latest version and it worked. Can this cause any issue here?

david26694 (changes requested):
Notebook looks super good, I'd also run the missing cells or delete them. Since you changed the name of the notebook, you need to change the reference to it in mkdocs.yaml. But after that is done, lgtm

Gabrielcidral1:
perfect, will do :)

Gabrielcidral1:
I think it's ready to go!

david26694 (approved):

PR #151, opened 2024-02-05
fix: sort values before the shift function

danielapinho:
Adding a sort before the shift, so we get the correct "last switch" from the dataset


david26694 (changes requested):
can you bump the version in setup.py, so a new release is done with the merge? 

also, we could have unit tests on dataframes that are not ordered by time col to test this behaviour, maybe not for this PR, but we can add an issue for a future PR

david26694 (approved):
lgtm, thanks!

PR #153, opened 2024-02-06
feat: Adding unit tests with unordered df

danielapinho:


david26694:
Merging this @danielapinho , although it fails in your local environment I think it's worth adding for a more robust test (and this is probably not the reason why it fails in your env anyway)

PR #154, opened 2024-02-21
Added washover usage example

pinsacco:


cluster_experiments/washover.py
```diff
@@ -43,8 +43,8 @@ def washover(
             df (pd.DataFrame): Input dataframe.
             truncated_time_col (str): Name of the truncated time column.
             treatment_col (str): Name of the treatment column.
-            cluster_cols (List[str]): List of clusters of experiment.
-            original_time_col (Optional[str], optional): Name of the original time column.
+            cluster_cols (List[str]): List of clusters of experiment. The list must include the 'truncated_time_col' as well.
+            original_time_col (str): Name of the original time column.
```
david26694, line 47:
```suggestion
original_time_col (Optional[str], optional): Name of the original time column.
```

david26694, line 47: can you revert this line to:
```python
original_time_col (Optional[str], optional): Name of the original time column.
```

cluster_experiments/washover.py
```diff
@@ -97,46 +97,66 @@ def washover(
...
+        there is a switch where the treatment is different.
 
         Args:
             df (pd.DataFrame): Input dataframe.
             truncated_time_col (str): Name of the truncated time column.
             treatment_col (str): Name of the treatment column.
-            cluster_cols (List[str]): List of clusters of experiment.
-            original_time_col (Optional[str], optional): Name of the original time column.
+            cluster_cols (List[str]): List of clusters of experiment. The list must include the 'truncated_time_col' as well.
+            original_time_col (str): Name of the original time column.
```
david26694, line 108: also revert this line:
original_time_col (Optional[str], optional): Name of the original time column.

david26694 (changes requested):
Last changes: I think it's easier to do an anti-join (https://stackoverflow.com/questions/38516664/anti-join-pandas) that is showing the rows that are in the  first df and not in the second, and this will be easier to visualize, the reader will see that they are only at the beginning of the hour. wdyt?
imagen

david26694:
(also, give the PR a better name pls)

david26694 (approved):

PR #156, opened 2024-02-26
Add washover input validation

pinsacco:


cluster_experiments/washover.py
```diff
@@ -10,6 +10,27 @@
...
+        cluster_cols: List[str],
+        original_time_col: str,
+    ):
+        if original_time_col not in df.columns:
+            raise ValueError(
+                "The original_time_col is not defined and/or not available in the dataframe columns."
+            )
+        if truncated_time_col not in cluster_cols:
+            raise ValueError(
+                f"The truncated_time_col '{truncated_time_col}' is not in the cluster columns."
```
david26694, line 26:
```suggestion
                f"{truncated_time_col = } is not in the cluster columns."
```
wdyt of something like this? and similar for original_time_col and cluster_cols

cluster_experiments/washover.py
```diff
@@ -10,6 +10,27 @@
...
 
+    def _validate_columns(
+        self,
+        df: pd.DataFrame,
+        truncated_time_col: str,
+        cluster_cols: List[str],
+        original_time_col: str,
+    ):
+        if original_time_col not in df.columns:
+            raise ValueError(
```
david26694, line 21: I'd add tests that validate that these conditions are indeed raising an error with pytest.raises. let me know if you need help with that

david26694: I'd also add a docstring to this method

david26694 (commented):
Left some comments

tests/splitter/test_washover.py
```diff
@@ -139,3 +140,44 @@ class Cfg:
         assert not washover_df.query("time >= '2022-01-01 01:00:00'").equals(
             out_df.query("time >= '2022-01-01 01:00:00'")
         )
+
+
+def test_truncated_time_not_in_cluster_cols():
+    msg = "is not in the cluster columns."
+    with pytest.raises(ValueError, match=msg):
+        df = pd.DataFrame(columns=["time_bin", "city", "time", "treatment"])
```
david26694, line 148: I'd create the dataframe before the with statement, I'd also add some comment on what the test tries to capture

david26694 (approved):
lgtm, left a small suggestion

david26694 (changes requested):
you also need to update setup.py to a newer version

david26694 (approved):
lgtm, great job

PR #157, opened 2024-02-28
fix: bump mkdocs to 1.4.0

david26694:


david26694:
fyi it works now @pinsacco , good catch

PR #158, opened 2024-02-29
fix: washover for cases with 0 non_time_cols

victorbr92:
Hey David! :wave:

Currently, if you try to add a washover for a switchback experiment where the only cluster is time, it fails with: `ValueError: No group keys passed!`
The test I proposed can reproduce the bug.

The solution is just to handle the case with non-grouping columns in the simplest way possible.

Let me know if you have any comments.


david26694 (changes requested):
hey victor! thanks for the PR, and good catch. The only change missing would be to update setup.py to do a new release of the library, to 0.14.0 for instance

victorbr92:
Ok, updated to 0.14.0

david26694 (approved):
Great!

PR #159, opened 2024-03-04
Update cluster_cols on random_splitter.py

danielapinho:
If the time_col is not provided as a cluster_col, add it after aggregating to switch length, to enable the assignment of treatment/control

### DIFF
--- a/cluster_experiments/random_splitter.py
+++ b/cluster_experiments/random_splitter.py
@@ -189,6 +189,8 @@ class SwitchbackSplitter(ClusteredSplitter):
         # Overwriting column, this is the worst! If we use the column as a covariate, we're screwed. Needs improvement
         df[_original_time_column(self.time_col)] = df[self.time_col]
         df[self.time_col] = self._get_time_col_cluster(df)
+        # add time_col if it is not already there
+        self.cluster_cols = list(set(self.cluster_cols + [self.time_col]))
         return df
 
     def assign_treatment_df(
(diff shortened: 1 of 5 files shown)


cluster_experiments/random_splitter.py
```diff
@@ -189,6 +189,8 @@ def _prepare_switchback_df(self, df: pd.DataFrame) -> pd.DataFrame:
         # Overwriting column, this is the worst! If we use the column as a covariate, we're screwed. Needs improvement
         df[_original_time_column(self.time_col)] = df[self.time_col]
         df[self.time_col] = self._get_time_col_cluster(df)
+        # add time_col if it is not already there
+        self.cluster_cols = list(set(self.cluster_cols + [self.time_col]))
```
david26694, line 193: I think it's better to do this in the init, I'd avoid changing attributes after the init if possible

david26694 (changes requested):

david26694:
Closing this because the following config will fail as shown in the test:
```python
    config = {
        "time_col": "time",
        "switch_frequency": "30min",
        "perturbator": "constant",
        "analysis": "ols_clustered",
        "splitter": "switchback_stratified",
        "cluster_cols": ["city"],
        "strata_cols": ["day_of_week", "hour_of_day", "city"],
        "target_col": "y",
        "n_simulations": 3,
    }

    power = PowerAnalysis.from_dict(config)
```

PR #167, opened 2024-03-09
feat: adding TwoEventsWashover base class and was modified validate c…

### LINKED ISSUE #146: Implement washover based on two events
Assume that:
* Let's say we have a switch at 12:00. 
* For each record in our experiment, we have two events' timestamps, login_timestamp and logout_timestamp. We want to apply washover such that if a record has login_timestamp in treatment X and logout_timestamp in treatment Y, we drop this row.

Manhe-dev:
base class and was modified validate columns

### DIFF
--- a/cluster_experiments/washover.py
+++ b/cluster_experiments/washover.py
@@ -286,6 +286,7 @@ class TwoEventsWashover(Washover):
             record_df (pd.DataFrame): Record dataframe.
             start_time_column (str): Name of the start time column in the record dataframe.
             end_time_column (str): Name of the end time column in the record dataframe.
+
         """
         self.calendar_df = calendar_df
         self.time_column_calendar = time_column_calendar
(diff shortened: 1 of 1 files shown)


cluster_experiments/washover.py
```diff
@@ -242,5 +242,124 @@ def from_config(cls, config) -> "Washover":
...
+        if start_time_column not in record_df.columns:
+            raise ValueError(
+                f"{start_time_column = } is not in the record dataframe columns and/or not specified as an input."
+            )
+        if end_time_column not in record_df.columns:
+            raise ValueError(
+                f"{end_time_column = } is not in the record dataframe columns and/or not specified as an input."
+            )
+
+    def washover(
```
david26694, line 315: I'd try to maintain the signature of the abstract washover class, if not it's very hard to put it into power analysis:
```python
    @abstractmethod
    def washover(
        self,
        df: pd.DataFrame,
        truncated_time_col: str,
        treatment_col: str,
        cluster_cols: List[str],
        original_time_col: Optional[str] = None,
    ) -> pd.DataFrame:
        """Abstract method to add washvover to the dataframe."""
```

david26694: For this reason, I'd send start and end columns to the init method

cluster_experiments/washover.py
```diff
@@ -242,5 +242,124 @@ def from_config(cls, config) -> "Washover":
...
+        |  4 | 11:01      | 12:01    |
+        |  5 | 12:01      | 12:05    |
+        +----+------------+----------+
+
+    This washover will drop record number 2.
+    """
+
+    def __init__(
+        self,
+        calendar_df: pd.DataFrame,
```
david26694, line 277: I'd try to avoid do to init with a calendar df, we've tried to keep constructors only based on very simple types like str, int and float. I'd assume the user already has the calendar in the record df, so you don't need to do a join, and it works like other washover classes

david26694 (commented):
Doing an (unsolicited) review, since I saw you started working on it and I'm going on holidays 🏖️

cluster_experiments/washover.py
Manhe-dev: okis

PR #168, opened 2024-04-10
Add synthetic control as analysis method

Gabrielcidral1:
As described by @david26694 we want to: "Add a wrapper to a synthetic control implementation that gives p-values, this should allow us to treat synthetic control as just another analysis method and check if it has higher power than simpler things"

This involved the creation of a new analysis class called SyntheticControlAnalysis, similar to the other types of analysis.  

However, in order to perform the analysis correctly, we need the pre experiment data as part of the fit to find the weights (fit_synthetic). In the implementation in main, PowerAnalysis class accepts a pre_experiment param, however this is only used for cuped purposes (where we add a column to df). Therefore, I had to create a new class called PowerAnalysisWithPreExperimentData, where the pre experiment df is also available. 

Furthermore, I had to create another splitter called FixedSizeClusteredSplitter, as we want only one cluster to be assigned as treatment and the rest as control. This was done to simplify the logic and to be more consistent with the usual application of synthetic control

This will not be implemented in this PR
- Allow power analysis with more than 1 treatment cluster
- Run it from power config
- Graphs on synthetics and donors 
- Parallel execution for p value calculation


david26694 (changes requested):

cluster_experiments/experiment_analysis.py
```diff
@@ -680,3 +683,181 @@ def analysis_point_estimate(self, df: pd.DataFrame, verbose: bool = False) -> fl
...
+    def analysis_pvalue(self, df: pd.DataFrame, verbose: bool = False) -> float:
+        """Returns the p-value of the analysis
+        Arguments:
+            df: dataframe containing the data to analyze
+            verbose (Optional): bool, prints the regression summary if True
+        """
+
+        # todo should I throw an error if more than one cluster column is passed?
+        control_pool = df[self.cluster_cols].iloc[:, 0].unique()
+        # control_pool = df[self.cluster_cols].iloc[:, 0].values
```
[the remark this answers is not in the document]
Gabrielcidral1: I will leave this for now because this is a working version with non parallel execution. Once the PR is ready for review (not draft) I will exclude any commented code

cluster_experiments/random_splitter.py
```diff
@@ -538,3 +537,32 @@ def from_config(cls, config):
             treatments=config.treatments,
             treatment_col=config.treatment_col,
         )
+
+
+class FixedTreatmentClustersSplitter(ClusteredSplitter):
+    def __init__(self, cluster_cols: List[str], n_treatment_clusters: int):
```
david26694, line 543: can you add some docstring explaining what this does and why is it different than other splitters?

cluster_experiments/experiment_analysis.py
```diff
@@ -680,3 +683,186 @@ def analysis_point_estimate(self, df: pd.DataFrame, verbose: bool = False) -> fl
...
+        ]
+
+        synthetic_donors = [
+            unit.assign(effect=unit[self.target_col] - unit["synthetic"])
+            for unit in synthetic_donors
+        ]
+
+        # first calculate the average effect after intervention for each unit, in a list comprehension
+        avg_effects = [
+            unit.query("treatment_period == 'After'")["effect"].mean()
```
david26694, line 805: I think it makes sense to add 'After' as an attribute of the class instead of being hardcoded

cluster_experiments/experiment_analysis.py
```diff
@@ -680,3 +683,186 @@ def analysis_point_estimate(self, df: pd.DataFrame, verbose: bool = False) -> fl
...
+        if not any(df[self.treatment_col] == 1):
+            raise ValueError("No treatment unit found in the data.")
+
+        y = df.query(f"{self.treatment_col}==1")[f"{self.target_col}"]  # treated
+
+        weights = get_w(X, y)
+
+        return weights
+
+    def synthetic_control(self, df: pd.DataFrame, treatment_unit) -> np.array:
```
david26694, line 771: what does this function actually do? I think it can be renamed to something more related to what it does

david26694: and I think it returns a dataframe, not an array

cluster_experiments/experiment_analysis.py
```diff
@@ -680,3 +683,186 @@ def analysis_point_estimate(self, df: pd.DataFrame, verbose: bool = False) -> fl
...
+            .values.dot(weights)
+        )
+
+        return df.query(
+            f"{''.join(self.cluster_cols)} == '{treatment_unit}'"
+        ).assign(  # add synthetic to treatment unit
+            synthetic=synthetic
+        )
+
+    def pvalue_based_on_hypothesis(
```
david26694, line 788: can you elaborate in a docstring what method do you use to obtain a pvalue?

david26694 (changes requested):

cluster_experiments/random_splitter.py
```diff
@@ -538,3 +537,46 @@ def from_config(cls, config):
             treatments=config.treatments,
             treatment_col=config.treatment_col,
         )
+
+
+class PredefinedTreatmentClustersSplitter(ClusteredSplitter):
```
david26694, line 542: wdyt about the name FixedSizeClusteredSplitter?

cluster_experiments/synthetic_control.py
```diff
@@ -0,0 +1,39 @@
+from functools import partial
+
+import numpy as np
+from numpy import ndarray
+from scipy.optimize import fmin_slsqp
+
+
+def loss_w(W: ndarray, X: ndarray, y: ndarray) -> float:
+    """
+    This function calculates the root mean square error (RMSE) between the actual and predicted values.
```
david26694, line 10: I'd mention that this is for a linear model

tests/splitter/test_predefined_treatment_clusters_splitter.py
```diff
@@ -0,0 +1,19 @@
+import pandas as pd
+
+from cluster_experiments.random_splitter import PredefinedTreatmentClustersSplitter
+
+
+def test_predefined_treatment_clusters_splitter():
```
david26694, line 6: can you add a test where you have more than 1 row per cluster?

tests/power_analysis/test_power_analysis_with_experiment_data.py
```diff
@@ -0,0 +1,55 @@
...
+    # Ensure 'date' column is of datetime type
+    df["date"] = pd.to_datetime(df["date"])
+
+    return df
+
+
+# Example usage
+
+
+def test_PowerAnalysisWithPreExperimentData():
```
david26694, line 35:
```suggestion
def test_power_analysis_with_pre_experiment_data():
```

tests/power_analysis/test_power_analysis_with_experiment_data.py
```diff
@@ -0,0 +1,55 @@
...
+    df = pd.DataFrame(combinations, columns=["user", "date"])
+    df["target"] = target_values
+
+    # Ensure 'date' column is of datetime type
+    df["date"] = pd.to_datetime(df["date"])
+
+    return df
+
+
+# Example usage
```
david26694, line 32:
```suggestion
```

tests/power_analysis/test_power_analysis_with_experiment_data.py
```diff
@@ -0,0 +1,55 @@
+from itertools import product
```
david26694, line 1: can we add the following two tests?
* AA test: check that if effect = 0, power = alpha, more or less
* Check that if effect is bigger, power is bigger

tests/analysis/test_synthetic_analysis.py
```diff
@@ -0,0 +1,83 @@
+# we want to create a test to confirm the synthetic control analysis works as expected. We will generate a some donors, and 2 of them will be double the target.
```
david26694, line 1: do you think it makes sense to drop this?

examples/short_example_synthetic_control.py
```diff
@@ -0,0 +1,58 @@
...
+        cluster_cols=["user"], time_col="date", intervention_date="2022-01-15"
+    )
+
+    pw = PowerAnalysisWithPreExperimentData(
+        perturbator=perturbator, splitter=sw, analysis=analysis, n_simulations=50
+    )
+
+    power = pw.power_analysis(df)
+    a = pw.power_line(df, average_effects=[0.1, 0.2, 0.5, 1, 1.5], n_jobs=-1)
+    set(pw.simulate_point_estimate(df))
```
david26694, line 57: why is this a set? let's print it and print(a) too no?

cluster_experiments/power_analysis.py
```diff
@@ -524,3 +524,65 @@ def check_inputs(self):
...
+
+
+class PowerAnalysisWithPreExperimentData(PowerAnalysis):
+    """
+    Same as PowerAnalysis, but adding pre-experiment param during split and perturbation. The PowerAnalysis class
+    accepts a pre_experiment param, however this is only used for cuped purposes (where we add a column to df). This
+    child class is used for cases where the pre experiment df is also available when the class is instantiated
+    """
+
+    def split(self, df: pd.DataFrame) -> pd.DataFrame:
```
david26694, line 536: label it as _split

cluster_experiments/power_analysis.py
```diff
@@ -524,3 +524,65 @@ def check_inputs(self):
...
+        treatment_df = treatment_df.query(
+            f"{self.treatment_col}.notnull()", engine="python"
+        ).query(
+            f"{self.treatment_col}.isin(['{self.treatment}', '{self.control}'])",
+            engine="python",
+        )
+
+        return treatment_df
+
+    def perturbate(
```
david26694, line 549: label it as _perturbate

cluster_experiments/power_analysis.py
```diff
@@ -524,3 +524,65 @@ def check_inputs(self):
...
+        self, treatment_df: pd.DataFrame, average_effect: Optional[float]
+    ) -> pd.DataFrame:
+
+        pre_experiment_df = treatment_df.query(
+            f"{self.analysis.time_col} <= '{self.analysis.intervention_date}'"
```
david26694, line 554: why not use the same sytanx as in the next statement? for consistency

cluster_experiments/power_analysis.py
```diff
@@ -524,3 +524,65 @@ def check_inputs(self):
...
+        treatment_df = treatment_df[
+            (treatment_df[self.analysis.time_col] > self.analysis.intervention_date)
+        ]
+
+        perturbed_df = self.perturbator.perturbate(
+            treatment_df, average_effect=average_effect
+        )
+
+        # get treatment cluster
+        treatment_cluster = perturbed_df.loc[
```
david26694, line 566: this method is a bit long, I'd try to:
* move part of the logic in here to another method
* unit-test it. this has quite a bit of logic and it's a good candidate to test that it's doing what is expected

cluster_experiments/experiment_analysis.py
```diff
@@ -680,3 +682,255 @@ def analysis_point_estimate(self, df: pd.DataFrame, verbose: bool = False) -> fl
...
+            pre_experiment_df.query(f"{self.treatment_col}==1")
+            .pivot(index=self.cluster_cols, columns=self.time_col)[self.target_col]
+            .T.iloc[:, 0]
+        )
+
+        weights = get_w(X, y, verbose)
+
+        return weights
+
+    def synthetic_control(
```
david26694, line 785: wdyt renaming to something like predict_synthetic?

cluster_experiments/experiment_analysis.py
```diff
@@ -680,3 +682,255 @@ def analysis_point_estimate(self, df: pd.DataFrame, verbose: bool = False) -> fl
...
+            unit.query(f"{self.time_col} >= '{self.intervention_date}'")[
+                "effect"
+            ].mean()
+            for unit in synthetic_donors
+        ]
+
+        treatment_index = np.where(
+            [df[self.treatment_col].any() for df in synthetic_donors]
+        )[0][0]
+        ate = avg_effects[treatment_index]  # this is the real treatment effect
```
david26694, line 853: I think a more natural way to return this is a dictionary with {cluster: effect} pairs. wdyt?

cluster_experiments/experiment_analysis.py
```diff
@@ -680,3 +682,255 @@ def analysis_point_estimate(self, df: pd.DataFrame, verbose: bool = False) -> fl
...
+
+        Returns:
+            tuple: A tuple containing the average treatment effect and a list of the placebo effects.
+        """
+        synthetic_donors = [
+            unit.assign(effect=unit[self.target_col] - unit["synthetic"])
+            for unit in synthetic_donors
+        ]
+
+        avg_effects = [
```
david26694, line 843: probably a dict comprehension will work better in here, see the comments below

david26694 (changes requested):
Super good! I'd also add a notebook where you show the effect is estimated in an unbiased way and showing some power lines

cluster_experiments/experiment_analysis.py
```diff
@@ -680,3 +682,255 @@ def analysis_point_estimate(self, df: pd.DataFrame, verbose: bool = False) -> fl
...
+                pre_experiment_df=pre_experiment_df,
+                verbose=verbose,
+            )
+            for cluster in clusters
+        ]
+
+        p_value = self.pvalue_based_on_hypothesis(synthetic_donors=synthetic_donors)
+        return p_value
+
+    def analysis_point_estimate(self, df: pd.DataFrame, verbose: bool = False) -> float:
```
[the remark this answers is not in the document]
Gabrielcidral1: I didn't reuse it because I would have to re declare the signature of analysis_point_estimate (so it accepts a treatment_cluster param). But I think it will indeed streamline the code

cluster_experiments/power_analysis.py
```diff
@@ -524,3 +524,65 @@ def check_inputs(self):
...
+            perturbed_df[self.treatment_col] == self.treatment,
+            self.analysis.cluster_cols[0],
+        ].unique()[0]
+
+        pre_experiment_df = pre_experiment_df.copy()
+
+        pre_experiment_df.loc[:, "treatment"] = np.where(
```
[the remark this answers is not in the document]
Gabrielcidral1: you're right, it's not doing anything

cluster_experiments/experiment_analysis.py
```diff
@@ -680,3 +682,226 @@ def analysis_point_estimate(self, df: pd.DataFrame, verbose: bool = False) -> fl
...
+
+    ```
+    """
+
+    def __init__(
+        self,
+        target_col: str = "target",
+        treatment_col: str = "treatment",
+        treatment: str = "B",
+        cluster_cols: Optional[List[str]] = None,
```
david26694, line 732: is this really optional?

david26694: if not, I suggest passing it as the first argument without a default

cluster_experiments/experiment_analysis.py
```diff
@@ -680,3 +682,226 @@ def analysis_point_estimate(self, df: pd.DataFrame, verbose: bool = False) -> fl
...
+        hypothesis: str = "two-sided",
+        time_col: str = "date",
+        intervention_date: str = None,
+    ):
+        super().__init__(cluster_cols)
```
david26694, line 737: can we init the super with more attributes? like target_col, treatment_col, etc

cluster_experiments/experiment_analysis.py
```diff
@@ -680,3 +682,226 @@ def analysis_point_estimate(self, df: pd.DataFrame, verbose: bool = False) -> fl
...
+        Returns the p-value of the analysis.
+        1. calculate the average effect after intervention for each unit.
+        2. count how many times the average effect is greater than the real treatment unit
+        3. Divide by the number of units. The result is the p-value using Fisher permutation test.
+        """
+
+        avg_effects = list(avg_effects.values())
+
+        if HypothesisEntries(self.hypothesis) == HypothesisEntries.LESS:
+            return np.mean(avg_effects < ate)
```
david26694, line 829: does this work with a list? I understood avg_effects needs to be an array

cluster_experiments/experiment_analysis.py
```diff
@@ -680,3 +682,226 @@ def analysis_point_estimate(self, df: pd.DataFrame, verbose: bool = False) -> fl
...
+        if HypothesisEntries(self.hypothesis) == HypothesisEntries.TWO_SIDED:
+            avg_effects = np.abs(avg_effects)
+            return np.mean(avg_effects > ate)
+
+        raise ValueError(f"{self.hypothesis} is not a valid HypothesisEntries")
+
+    def _get_treatment_cluster(self, df: pd.DataFrame) -> str:
+        """Returns the treatment cluster"""
+        treatment_df = df[df[self.treatment_col] == 1]
+
```
david26694, line 841:
```suggestion
```

cluster_experiments/experiment_analysis.py
```diff
@@ -680,3 +682,226 @@ def analysis_point_estimate(self, df: pd.DataFrame, verbose: bool = False) -> fl
...
+        treatment_cluster = self._get_cluster_column(treatment_df).unique()[0]
+        return treatment_cluster
+
+    def analysis_pvalue(self, df: pd.DataFrame, verbose: bool = False) -> float:
+        """
+        Calculate the p-value using the treatment effect calculated from analysis_point_estimate
+        for the actual treatment cluster and comparing it with placebo effects from other clusters.
```
david26694, line 848: I'd mention that you're using a permutation test (are you?)

cluster_experiments/experiment_analysis.py
```diff
@@ -680,3 +682,226 @@ def analysis_point_estimate(self, df: pd.DataFrame, verbose: bool = False) -> fl
...
+        )
+
+        df["effect"] = df[self.target_col] - df["synthetic"]
+        avg_effect = df["effect"].mean()
+        return avg_effect
+
+    def _split_pre_experiment_df(self, df: pd.DataFrame):
+        """Split the dataframe into pre-experiment and experiment dataframes"""
+        pre_experiment_df = df[(df[self.time_col] <= self.intervention_date)]
+
```
david26694, line 893: feel free to reject
```suggestion
```

cluster_experiments/experiment_analysis.py
```diff
@@ -680,3 +682,226 @@ def analysis_point_estimate(self, df: pd.DataFrame, verbose: bool = False) -> fl
...
+        df = df[(df[self.time_col] > self.intervention_date)]
+
```
david26694, line 895: feel free to reject
```suggestion
```

cluster_experiments/experiment_analysis.py
```diff
@@ -680,3 +682,226 @@ def analysis_point_estimate(self, df: pd.DataFrame, verbose: bool = False) -> fl
...
+    from cluster_experiments.experiment_analysis import SyntheticControlAnalysis
+    import pandas as pd
+    import numpy as np
+
+    dates = pd.date_range(start_date, end_date, freq="d")
+
+    users = [f"User {i}" for i in range(N)]
+
+    # Create a combination of each date with each user
+    combinations = list(product(users, dates))
```
david26694, line 712: do you need to import product in here?

cluster_experiments/experiment_analysis.py
```diff
@@ -680,3 +682,226 @@ def analysis_point_estimate(self, df: pd.DataFrame, verbose: bool = False) -> fl
...
+        self.treatment = treatment
+        self.treatment_col = treatment_col
+        self.cluster_cols = cluster_cols
+        self.hypothesis = hypothesis
+        self.time_col = time_col
+        self.intervention_date = intervention_date
+
+    # todo  check is that time_col not in cluster_cols
+
+    def get_pvalue(self, df: pd.DataFrame) -> float:
```
david26694, line 749: why do you need to inherit this? isn't this the same as the parent method?

cluster_experiments/experiment_analysis.py
```diff
@@ -680,3 +682,226 @@ def analysis_point_estimate(self, df: pd.DataFrame, verbose: bool = False) -> fl
...
+
+        y = (
+            pre_experiment_df.query(f"{self.treatment_col}==1")
+            .pivot(index=self.cluster_cols, columns=self.time_col)[self.target_col]
+            .T.iloc[:, 0]
+        )
+
+        weights = get_w(X, y, verbose)
+
+        return weights
```
david26694, line 780: is this actually a list? my pylint is complaining

cluster_experiments/experiment_analysis.py
```diff
@@ -680,3 +682,226 @@ def analysis_point_estimate(self, df: pd.DataFrame, verbose: bool = False) -> fl
...
+
+    def __init__(
+        self,
+        target_col: str = "target",
+        treatment_col: str = "treatment",
+        treatment: str = "B",
+        cluster_cols: Optional[List[str]] = None,
+        hypothesis: str = "two-sided",
+        time_col: str = "date",
+        intervention_date: str = None,
```
david26694, line 735: type hint is wrong in here. None is not str

david26694: I'd add it as positional arg, not kwarg

cluster_experiments/experiment_analysis.py
```diff
@@ -680,3 +682,226 @@ def analysis_point_estimate(self, df: pd.DataFrame, verbose: bool = False) -> fl
...
+        return self.analysis_pvalue(df=df)
+
+    def fit_synthetic(self, pre_experiment_df: pd.DataFrame, verbose: bool) -> list:
+        """Returns the weights of each donor"""
+
+        if not any(pre_experiment_df[self.treatment_col] == 1):
+            raise ValueError("No treatment unit found in the data.")
+
+        X = (
+            pre_experiment_df.query(f"{self.treatment_col}==0")
```
david26694, line 767:
```suggestion
            pre_experiment_df.query(f"{self.treatment_col} == 0")
```

cluster_experiments/experiment_analysis.py
```diff
@@ -680,3 +682,226 @@ def analysis_point_estimate(self, df: pd.DataFrame, verbose: bool = False) -> fl
...
+            .pivot(index=self.cluster_cols, columns=self.time_col)[self.target_col]
+            .T
+        )
+
+        y = (
+            pre_experiment_df.query(f"{self.treatment_col}==1")
```
david26694, line 773:
```suggestion
            pre_experiment_df.query(f"{self.treatment_col} == 1")
```

cluster_experiments/experiment_analysis.py
```diff
@@ -680,3 +682,226 @@ def analysis_point_estimate(self, df: pd.DataFrame, verbose: bool = False) -> fl
...
+            df[self._get_cluster_column(df) != treatment_cluster]
+            .pivot(index=self.time_col, columns=self.cluster_cols)[self.target_col]
+            .values.dot(weights)
+        )
+
+        return df[self._get_cluster_column(df) == treatment_cluster].assign(
+            synthetic=synthetic
+        )  # add synthetic to treatment cluster
+
+    def pvalue_based_on_hypothesis(self, ate: float, avg_effects: dict) -> float:
```
david26694, line 818: avg_effects is actually Dict[str, float], no?

cluster_experiments/experiment_analysis.py
```diff
@@ -680,3 +682,226 @@ def analysis_point_estimate(self, df: pd.DataFrame, verbose: bool = False) -> fl
...
+            .T.iloc[:, 0]
+        )
+
+        weights = get_w(X, y, verbose)
+
+        return weights
+
+    def fit_predict_synthetic(
+        self,
+        pre_experiment_df: pd.DataFrame,
```
david26694, line 784: any reason why pre_df is first argument? for consistency with the library, I'd add df as first argument

cluster_experiments/experiment_analysis.py
```diff
@@ -680,3 +682,226 @@ def analysis_point_estimate(self, df: pd.DataFrame, verbose: bool = False) -> fl
...
+        df: pd.DataFrame,
+        treatment_cluster,
```
david26694, line 786: type hint missing (do we know the type?)

cluster_experiments/power_analysis.py
```diff
@@ -524,3 +524,43 @@ def check_inputs(self):
         self.check_target_col()
         self.check_treatment()
         self.check_clusters()
+
+
+class PowerAnalysisWithPreExperimentData(PowerAnalysis):
```
david26694, line 529: I'd also override check_inputs, by checking that  isinstance(self.analysis, SyntheticControlAnalysis). or that hasattr(self.analysis, _split_pre_experiment_df) or something like this.

cluster_experiments/random_splitter.py
```diff
@@ -538,3 +537,46 @@ def from_config(cls, config):
...
+
+    Attributes:
+        cluster_cols (List[str]): List of columns to use as clusters.
+        n_treatment_clusters (int): The predefined number of treatment clusters.
+
+    """
+
+    def __init__(self, cluster_cols: List[str], n_treatment_clusters: int):
+        self.n_treatment_clusters = n_treatment_clusters
+        super().__init__(cluster_cols=cluster_cols)
```
david26694, line 556: can we reorder the two lines? the common thing is to do super first

cluster_experiments/synthetic_control_utils.py
```diff
@@ -0,0 +1,40 @@
...
+
+    Parameters:
+    W (numpy.ndarray): The weights vector used for predictions.
+    X (numpy.ndarray): The input data matrix.
+    y (numpy.ndarray): The actual output vector.
+
+    Returns:
+    float: The calculated RMSE.
+    """
+
```
david26694, line 22:
```suggestion
```

examples/short_example_synthetic_control.py
```diff
@@ -0,0 +1,56 @@
...
+    )
+
+    pw = PowerAnalysisWithPreExperimentData(
+        perturbator=perturbator, splitter=sw, analysis=analysis, n_simulations=50
+    )
+
+    power = pw.power_analysis(df)
+    print(f"{power = }")
+    print(pw.power_line(df, average_effects=[0.1, 0.2, 0.5, 1, 1.5], n_jobs=-1))
+    print(pw.simulate_point_estimate(df))
```
david26694, line 56: adding it here because can't add in notebook. you have an extra parenthesis in this line:
```python
pl = pw.power_line(df, average_effects=[0.1, 0.2, 0.5, 1, 1.5], n_jobs=-1))
```

david26694: also, in the notebook we don't seem to have power = alpha for effect = 0. do you know why might this happen?

cluster_experiments/experiment_analysis.py
david26694: yep, this is not failing because you need to add the class in test_docs.py. You also need to add it init in the cluster_experiments folder for it to be easier to import, and in readme for it to be more visible.

examples/short_example_synthetic_control.py
david26694: can you also add some simulations of point estimates in the notebook? so users can see that the estimate is unbiased?

cluster_experiments/synthetic_control_utils.py
david26694: this function is easily testable; I'd test it with some hardcoded values

cluster_experiments/experiment_analysis.py
```diff
@@ -680,3 +682,226 @@ def analysis_point_estimate(self, df: pd.DataFrame, verbose: bool = False) -> fl
...
+            )
+            for cluster in clusters
+        }
+
+        ate = synthetic_donors[treatment_cluster]
+        synthetic_donors.pop(treatment_cluster)
+
+        return self.pvalue_based_on_hypothesis(ate=ate, avg_effects=synthetic_donors)
+
+    def analysis_point_estimate(
```
david26694, line 868: can you add a test for this? add a very big effect and check that the point estimate is similar

cluster_experiments/experiment_analysis.py
```diff
@@ -680,3 +682,255 @@ def analysis_point_estimate(self, df: pd.DataFrame, verbose: bool = False) -> fl
...
+        df["effect"] = df[self.target_col] - df["synthetic"]
+        avg_effect = df["effect"].mean()
+
+        return avg_effect
+
+    @classmethod
+    def from_config(cls, config):
+        """Creates an OLSAnalysis object from a PowerConfig object"""
+        return cls(
+            target_col=config.target_col,
```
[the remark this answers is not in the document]
david26694: it's just if you want this to be able to create from dict we need to work a little on this. do you think it makes sense to init from dict? or the class init is good enough?

cluster_experiments/power_analysis.py
```diff
@@ -524,3 +524,43 @@ def check_inputs(self):
...
+            engine="python",
+        )
+
+        return treatment_df
+
+    def _perturbate(
+        self, treatment_df: pd.DataFrame, average_effect: Optional[float]
+    ) -> pd.DataFrame:
+
+        df, pre_experiment_df = self.analysis._split_pre_experiment_df(treatment_df)
```
david26694, line 553: pylint is failing because analysis doesn't always have _split_pre_experiment_df. I'm thinking how to address this. one option is following the comment above, but it won't solve the pylint issue. another one is adding a _split_pre_experiment_df method in the abstractclass that does raise NotImplementedError

david26694 (changes requested):
Supper good progress, I added some more comments

david26694:
in the notebook, I think it'd be cool to compare power lines and point estimate distrubtions of clusteredOLS and synthetic control

cluster_experiments/power_analysis.py
```diff
@@ -524,3 +524,43 @@ def check_inputs(self):
...
+
+        perturbed_df = self.perturbator.perturbate(df, average_effect=average_effect)
+
+        return pd.concat([perturbed_df, pre_experiment_df])
+
+    def _split_and_perturbate(
```
david26694, line 559: I'd consider moving this in the parent class, remove the _split method in this class, and just override the _perturbate

cluster_experiments/experiment_analysis.py
Gabrielcidral1: Maybe we change it to:

cluster_cols: Optional[List[str]] = ['cluster']

so we keep the idea of always having a default

david26694: wdyt about `cluster_cols: Union[List[str], Tuple[str]] = ('cluster',)`
and then doing `self.cluster_cols = list(cluster_cols)`?

examples/short_example_synthetic_control.py
Gabrielcidral1: Is this true even when the groups aren't equal? I have some weird results using OLS for 1 cluster treatment and the rest in control (power around 70%)

cluster_experiments/experiment_analysis.py
Gabrielcidral1: yep it does. Actually avg_effects comes as a dict, but I had to convert to list to run `np.mean(avg_effects < ate)` . I guess I could have converted it to array as well

examples/short_example_synthetic_control.py
david26694: ah, maybe the variance explodes with 1 treatment cluster

cluster_experiments/experiment_analysis.py
david26694: this is what I'm not following:
```python
>>> import numpy as np
>>> np.mean([1, 2] < 1)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: '<' not supported between instances of 'list' and 'int'
```

examples/short_example_synthetic_control.py
david26694: if you try:
```python
sw = FixedSizeClusteredSplitter(n_treatment_clusters=5, cluster_cols=["user"])
```
for ols, do you have the same issue?

david26694: btw does synthethic only work with n_treatment_clusters=1?

cluster_experiments/experiment_analysis.py
Gabrielcidral1: then I prefer to leave it empty haha

cluster_experiments/experiment_analysis.py
Gabrielcidral1: it works because the ate is a np.float

image

examples/short_example_synthetic_control.py
Gabrielcidral1: yes, it only work with 1 treatment cluster. I couldn't figure out an easy way to run several treatment clusters. 

Basically,  we need to find the best combination of weights for 1 treatment cluster at a time. So this means we need to run the model sequentially, for every treatment cluster. However, we need to store somewhere which are the treatment cluster because they can't be used as donors.

cluster_experiments/experiment_analysis.py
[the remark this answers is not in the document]
david26694: for an issue in the future

cluster_experiments/power_analysis.py
```diff
@@ -273,11 +273,9 @@ def power_analysis(
...
-    def _split_and_perturbate(
-        self, df: pd.DataFrame, average_effect: Optional[float]
-    ) -> pd.DataFrame:
+    def _split(self, df: pd.DataFrame) -> pd.DataFrame:
         """
-        Split and perturbate dataframe.
+        Split dataframe.
         Args:
             df: Dataframe with outcome variable
             average_effect: Average effect of treatment. If None, it will use the perturbator average effect.
```
david26694, line 281: drop avg effect in here

cluster_experiments/power_analysis.py
```diff
@@ -290,11 +288,33 @@ def _split_and_perturbate(
...
             engine="python",
         )
+
+        return treatment_df
+
+    def _perturbate(
+        self, treatment_df: pd.DataFrame, average_effect: Optional[float]
+    ) -> pd.DataFrame:
+        """
+        Split and perturbate dataframe.
```
david26694, line 298:
```suggestion
        Perturbate dataframe using perturbator.
```

cluster_experiments/synthetic_control_utils.py
```diff
@@ -0,0 +1,59 @@
...
+        partial(loss_w, X=X, y=y),
+        w_start,
+        f_eqcons=lambda x: np.sum(x) - 1,
+        bounds=bounds,
+        disp=verbose,
+    )
+    return weights
+
+
+def generate_data(N, start_date, end_date):
```
david26694, line 42: why is this function needed in the library?

examples/short_example_synthetic_control.py
```diff
@@ -0,0 +1,26 @@
+from cluster_experiments.experiment_analysis import SyntheticControlAnalysis
```
david26694, line 1: some comments about the jupyter:
* point estimate is super good
* power has many mentions to OLS but I don't see it in the plot
* power goes from 0 to 1, can you pick some effects in the middle?

setup.py
```diff
@@ -39,7 +39,7 @@
     "mkdocstrings==0.18.0",
     "jinja2<3.1.0",
     "mkdocs-jupyter==0.22.0",
-    "plotnine==0.8.0",
+    "plotnine>=0.8.0",
     "matplotlib>=3.4.3",
 ]
```
david26694, line 44: bump version of cluster-experiments to 0.16.0, I did a release that moved it to 0.15.0

cluster_experiments/synthetic_control_utils.py
david26694: I see it being used in a couple places:
* short example -> if it were only here, I'd add it examples/
* unit tests -> if it were only here, I'd add it tests/

since used in two it makes sense to keep it, but I'd make it internal, like `_generate_data`, and change the name to something more specific: `_generate_synthetic_control_data`

tests/power_analysis/test_power_analysis_with_pre_experiment_data.py
```diff
@@ -0,0 +1,76 @@
...
+import numpy as np
+import pandas as pd
+
+from cluster_experiments.experiment_analysis import SyntheticControlAnalysis
+from cluster_experiments.perturbator import ConstantPerturbator
+from cluster_experiments.power_analysis import PowerAnalysisWithPreExperimentData
+from cluster_experiments.random_splitter import FixedSizeClusteredSplitter
+
+
+def generate_data(N, start_date, end_date):
```
david26694, line 12: if you had generate_data before, do you need this?

cluster_experiments/power_analysis.py
```diff
@@ -527,6 +547,24 @@ def check_inputs(self):
         self.check_clusters()
 
 
+class PowerAnalysisWithPreExperimentData(PowerAnalysis):
+    """
+    Same as PowerAnalysis, but allowing a perturbation only at experiment period and keeping pre-experiment df intact. The PowerAnalysis class
```
david26694, line 552: I'd mention here that this is supposed to work mainly for diff-in-diff or synthetic control-like estimators.

cluster_experiments/power_analysis.py
```diff
@@ -527,6 +547,24 @@ def check_inputs(self):
         self.check_clusters()
 
 
+class PowerAnalysisWithPreExperimentData(PowerAnalysis):
```
david26694, line 550: I'm thinking about what to do to protect the user to use this in the wrong cases. Perhaps we should add a check like
`hasattr(self.analysis, "_split_pre_experiment_df")` (and raise error if false)?

Or derive the synethetic control from PreExperimentAnalysis
```
class SyntheticControlAnalysis(ExperimentAnalysis, PreExperimentAnalysis):
```

and do something like `isinstance(self.analysis, PreExperimentAnalysis)`, and raise error if false.

wdyt?

cluster_experiments/experiment_analysis.py
```diff
@@ -740,3 +747,231 @@ def analysis_standard_error(self, df: pd.DataFrame, verbose: bool = False) -> fl
...
+
+        ate = synthetic_donors[treatment_cluster]
+        synthetic_donors.pop(treatment_cluster)
+
+        return self.pvalue_based_on_hypothesis(ate=ate, avg_effects=synthetic_donors)
+
+    def analysis_point_estimate(
+        self,
+        df: pd.DataFrame,
+        treatment_cluster: Optional[str] = "",
```
david26694, line 943:
```suggestion
        treatment_cluster: str = "",
```

david26694: either this or:
```
        treatment_cluster: Optional[str] = None
```

david26694: (I prefer the second and then `if treatment_cluster is None`)

cluster_experiments/experiment_analysis.py
```diff
@@ -740,3 +747,231 @@ def analysis_standard_error(self, df: pd.DataFrame, verbose: bool = False) -> fl
...
+
+        if treatment_cluster == "":
+            treatment_cluster = self._get_treatment_cluster(df)
+
+        df = self.fit_predict_synthetic(
+            df, pre_experiment_df, treatment_cluster, verbose=verbose
+        )
+
+        df["effect"] = df[self.target_col] - df["synthetic"]
+        avg_effect = df["effect"].mean()
```
david26694, line 959: I'd put a comment that you're averaging accross the time windows

david26694 (changes requested):
one last review, added some minor comments

cluster_experiments/synthetic_control_utils.py
Gabrielcidral1: I don't get it, if I keep it in utils and rename with underscore, then it would be wrong to import to other modules right? 

Do you mean to remove from modules and have 2 different definitions at short examples and unit tests?

cluster_experiments/power_analysis.py
Gabrielcidral1: Yes I agree that this can be confusing. What I am thinking now is to rename to something more related to Synthetic control (PowerAnalysisSyntheticControl). Once another estimator uses this class than we can think again of a better naming. 

The best solution would be to wrap both into one class. The user instantiate the class and the behavior will depend on the analysis

cluster_experiments/synthetic_control_utils.py
david26694: you're right, I don't have much control on the use of underscores, I was thinking adding it because it's not something that is useful to users (users should provide their data), but I agree that I'm skipping some standard rules when importing it. then never mind about underscore, just rename it to something like generate_synthetic_control_data

cluster_experiments/power_analysis.py
Gabrielcidral1: yep makes sense

cluster_experiments/__init__.py
david26694, line 1: Need to resolve the conflict with master, this is not importing NormalPowerAnalysis. also, your new class is missing from here

cluster_experiments/power_analysis.py
```diff
@@ -530,6 +549,29 @@ def check_inputs(self):
...
```
david26694, line 552: issues with the conflict with master, needs to copy normalpoweranalysis

cluster_experiments/experiment_analysis.py
```diff
@@ -740,3 +747,231 @@ def analysis_standard_error(self, df: pd.DataFrame, verbose: bool = False) -> fl
...
+        return avg_effect
+
+    def _split_pre_experiment_df(self, df: pd.DataFrame):
+        """Split the dataframe into pre-experiment and experiment dataframes"""
+        pre_experiment_df = df[(df[self.time_col] <= self.intervention_date)]
+        df = df[(df[self.time_col] > self.intervention_date)]
+        return df, pre_experiment_df
+
+    @classmethod
+    def from_config(cls, config):
```
david26694, line 969: let's drop this method for now, since it is not to be used

README.md
```diff
@@ -169,6 +169,8 @@ print(f"{power = }")
 The library offers the following classes:
 
 * Regarding power analysis:
+    * `PowerAnalysis`: to run power analysis on a clustered/switchback design
+    * `PowerAnalysisWithPreExperimentData`: to run power analysis on a clustered/switchback design, but adding pre-experiment df during split and perturbation (especially useful for Synthetic Control)
     * `PowerAnalysis`: to run power analysis on any experiment design, using simulation
```
david26694, line 174:
```suggestion
```

README.md
```diff
@@ -169,6 +169,8 @@ print(f"{power = }")
 The library offers the following classes:
 
 * Regarding power analysis:
+    * `PowerAnalysis`: to run power analysis on a clustered/switchback design
```
david26694, line 172:
```suggestion
    * `PowerAnalysis`: to run power analysis on any experiment design, using simulation
```

mkdocs.yml
```diff
@@ -23,6 +23,8 @@ nav:
   - Normal Power:
       - Compare with simulation: normal_power.ipynb
       - Time-lines: normal_power_lines.ipynb
+  - Synthetic control: synthetic_control.ipynb
+  - Normal Power: normal_power.ipynb
```
david26694, line 27:
```suggestion
```

setup.py
```diff
@@ -64,7 +64,6 @@
         "Programming Language :: Python :: 3.9",
         "Programming Language :: Python :: 3.10",
         "Programming Language :: Python :: 3.11",
```
david26694, line 66:
```suggestion
        "Programming Language :: Python :: 3.11",
        "Programming Language :: Python :: 3.12",
```

david26694:
(last suggestions and we merge)

david26694 (approved):
amazing job as always :)

PR #169, opened 2024-04-10
[94] feat: create exact power calculation class

### LINKED ISSUE #94: Create exact power calculation class
A new PowerAnalysis class needs to be set up. Something like:
`class ExactPowerAnalysis`, that implements the power_analysis and power_line methods using the linear model formula. 

I have this script I used a long time ago:

```python

import numpy as np
import pandas as pd
from cluster_experiments import PowerAnalysis
from scipy.stats import norm

def power_2_tails(df, splitter, alpha, ate, regressor):
    """Power of a test with a given alpha, ate and sigma

    Parameters
    ----------
    alpha : float
        Significance level
    ate : float
        Average treatment effect
    sigma : float
        Standard deviation of the difference of means (already divided by n)
    """
    df_treated = splitter.assign_treatment_df(df).assign(
        treatment=lambda x: (x.treatment == 'B').astype(int),
    )
    import statsmodels.api as sm
    fitted_ols = sm.OLS.from_formula(f"target ~ treatment + {regressor}", data=df_treated).fit()

    # get standard error of the regression coefficient
    se = fitted_ols.bse["treatment"]

    from scipy.stats import norm
    z_alpha = norm.ppf(1 - alpha / 2) 
    norm_cdf = norm.cdf(z_alpha - ate / se)
    norm_cdf_2 = norm.cdf(-z_alpha - ate / se)
    return 1 - norm_cdf + norm_cdf_2


# Create fake data
N = 2_000
alpha = 0.05
average_effect = 0.5
sigma = 1
df = pd.DataFrame(
    {
        "target": np.random.normal(0, sigma, size=2 * N),
        "regressor": np.random.normal(0, sigma, size=2 * N),
        "better_regressor": np.random.normal(0, sigma, size=2 * N),
    }
).assign(
    target=lambda x: x.target + x.regressor * 2 + x.better_regressor * 10
)


config = {
    "analysis": "ols_non_clustered",
    "perturbator": "uniform",
    "splitter": "non_clustered",
    "n_simulations": 1000,
    "covariates": ["regressor"],
    "alpha": alpha
}
pw = PowerAnalysis.from_dict(config)

pw_better = PowerAnalysis.from_dict({
    **config,
    "covariates": ["better_regressor"],
})


EFFECTS = [0, 0.1, 0.2, 0.3, 0.4, 0.5]


powers = pw.power_line(df, average_effects=EFFECTS)

powers_better = pw_better.power_line(df, average_effects=EFFECTS)

powers_exact = {}
powers_exact_better = {}
for average_effect in EFFECTS:
    powers_exact[average_effect] = power_2_tails(df, pw.splitter, alpha, average_effect, "regressor")
    powers_exact_better[average_effect] = power_2_tails(df, pw.splitter, alpha, average_effect, "better_regressor")



powers_better, powers
powers_exact_better, powers_exact


import matplotlib.pyplot as plt


plt.plot(EFFECTS, powers.values(), label="regressor")
plt.plot(EFFECTS, powers_better.values(), label="better regressor")
plt.plot(EFFECTS, powers_exact.values(), label="exact regressor")
plt.plot(EFFECTS, powers_exact_better.values(), label="exact better regressor")
plt.legend()
plt.xlabel("Average effect")
plt.ylabel("Power")
plt.title("Power of the test")
plt.show()


```

to compare exact vs simulation.

ludovico-lanni:
resolves #94


cluster_experiments/power_analysis.py
```diff
@@ -428,99 +580,73 @@ def from_config(cls, config: PowerConfig) -> "PowerAnalysis":
...
+                hypothesis=hypothesis,
+            )
+
+            self.check_inputs()
+
+        def power_analysis(
+            self,
+            df: pd.DataFrame,
+            pre_experiment_df: Optional[pd.DataFrame] = None,
+            average_effect: Optional[float] = None,
```
ludovico-lanni, line 640: need to change to required

ludovico-lanni:
TODO: add docstrings for AAPowerAnalysis class
TODO: add unit tests

@david26694 does the abstract interface make sense to you?

cluster_experiments/power_analysis.py
```diff
@@ -428,99 +580,73 @@ def from_config(cls, config: PowerConfig) -> "PowerAnalysis":
...
-
-        has_time_col = hasattr(self.splitter, "time_col")
-        assert not (
-            has_time_col
-            and has_splitter_clusters
-            and self.splitter.time_col not in self.splitter.cluster_cols
-        ), "in switchback splitters, time_col must be in cluster_cols"
-
     def check_inputs(self):
         self.check_covariates()
```
david26694, line 598: here you can call super, no?

cluster_experiments/power_analysis.py
```diff
@@ -105,26 +262,21 @@ def __init__(
...
             features_cupac_model=features_cupac_model,
+            seed=seed,
+            hypothesis=hypothesis,
         )
-        if seed is not None:
-            random.seed(seed)  # seed for splitter
-            np.random.seed(seed)  # seed for the binary perturbator
-            # may need to seed other stochasticity sources if added
+        self.perturbator = perturbator
+        self.n_simulations = n_simulations
```
david26694, line 279: I think n_simulations should be passed to the AA too, so I would move it to the parent

cluster_experiments/power_analysis.py
david26694: why?

david26694 (changes requested):
added some minor comments

cluster_experiments/power_analysis.py
ludovico-lanni: do you mean using the check_inputs method of super, and then add the additional check_perturbator_consistency() method to overwrite the subclass check_inputs() method?

cluster_experiments/power_analysis.py
ludovico-lanni: hmmm I'm taking out the perturbator as the power analysis is based on a deterministic formula. The Optional I guess was due to the fact that you can specify the effect in the perturbator, as far as I saw in the comments through the code. Without perturbator, does it make sense to have it as optional? How can I get the power value if I pass no effect?

cluster_experiments/power_analysis.py
ludovico-lanni: ok! was not aware of this :) In that case if n>1, the power for a given effect should be aggregated like mean(), reflecting the randomness in the split?

cluster_experiments/power_analysis.py
david26694: ah, you're right

cluster_experiments/power_analysis.py
david26694: the idea is averaging standard error, not the power, thuogh I'm not sure if they are equivalent

cluster_experiments/power_analysis.py
david26694: something like:
```python
super.check_inputs()
self.check_perturbator_consistency()
```
though my syntax may be off

david26694:
hey @ludovico-lanni , addressed your comments. on the name of the class, I'm not 100% convinced haha

cluster_experiments/experiment_analysis.py
```diff
@@ -316,6 +374,18 @@ def analysis_point_estimate(self, df: pd.DataFrame, verbose: bool = False) -> fl
...
+    def analysis_standard_error(self, df: pd.DataFrame, verbose: bool = False) -> float:
+        """Returns the standard error of the analysis
+        Arguments:
+            df: dataframe containing the data to analyze
+            verbose (Optional): bool, prints the regression summary if True
+        """
+        results_ols = sm.OLS.from_formula(
+            self.formula,
+            data=df,
+        ).fit()
```
ludovico-lanni, line 386: You forgot to add the `cov_type` and `cov_kwds` so you are getting the std_err assuming independence between all samples

ludovico-lanni (commented):
Hey @david26694 ! Thanks a lot for picking this up! I had a quick look and I'm quite sure I spotted the problem behind the failed comparison of the two power methods for clustered OLS. It is on the way you are computing the std_err, as it seems the specification of the clustered covariance structure is missing

david26694 (approved):

PR #192, opened 2024-11-19
feat: Delta Method ratio metric w/CUPED

LGonzalezGomez:
Added CUPED, tests and A-A comparison to illustrate the importance of using Delta Method instead than a naive t-test or normal OLS.


LGonzalezGomez:
Need to refactor code and remove code duplication. 
Might create utils for CUPED similar to that of CUPAC in the library.

PR #193, opened 2024-11-21
142 add delta method as analysis method

david26694:


LGonzalezGomez:
@david26694 I created this other [PR](https://github.com/david26694/cluster-experiments/pull/194) with a cleaner git history. Also now it should be working in tandem with CUPAC out-of-the-box.
I would close this PR. Merge first Aurelio's PR and then the other one once we finished

PR #194, opened 2024-11-22
feat: add cupac to Delta Method

LGonzalezGomez:
This PR add Delta Method with variance reduction. It also is now working with PowerAnalysis pipelines out-of-the-box.

Also added a couple of unit tests to ensure it is working appropiately.

### DIFF
--- a/README.md
+++ b/README.md
@@ -312,6 +312,7 @@ The library offers the following classes:
     * `HypothesisTest`: to define a Hypothesis Test with a metric, analysis method, optional analysis configuration, and optional dimensions
     * `AnalysisPlan`: to define a plan of analysis with a list of Hypothesis Tests for a dataset and the experiment variants. The `analyze()` method runs the analysis and returns the results
     * `AnalysisResults`: to store the results of the analysis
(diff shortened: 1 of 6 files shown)


LGonzalezGomez:
Some design consideration will shortly be revisited. As there is boilerplate that can be cleaned up.
Also need to write the notebook as example of usage and for mkdocs

LGonzalezGomez:
Would like to discuss the approach used. Using as an example CVR metric the steps are:

1. Fit model at session level if there will be a conversion or not. We use pre-experiment data for this. We only fit the numerator as we are doing this at session level still.
2. Predict conversions at session level on analysis_df
3. Aggregate predictions at cluster level (i.e. user) so that we have the needed covariate.
4. Apply Delta Method on target metric CVR and CUPAC covariate with is a prediction CVR_predicted = predicted_conversions/num_sessions. 
5. Do variance reduction following CUPED estimator, calculating the apropiate thetas
6. Compute variance after linearizing the ratio metrics with the Delta Method

This should IMO work in general if the procedure is correct. Happy to discuss. If this is a good approach, I think the code is cleaner and we don't have to do any type of overrides or nasty stuff.
Happy to hear feedback here!

PR #195, opened 2024-11-22
Move to uv as package manager

LGonzalezGomez:
Since we are in 2024 we might as well go to uv directly. I created a toml but we should check config.


Makefile
```diff
@@ -15,20 +15,19 @@ coverage_xml:
...
-	python -m pip install -e ".[dev]"
+	python -m pip install uv
+	uv sync --extra "dev"
 	pre-commit install
 
 install-test:
 	pip install --upgrade pip setuptools wheel
-	python -m pip install -e ".[test]"
-
-install-only-test:
```
LGonzalezGomez, line 29: @david26694 Is this needed? Seems weird to me and is a bit weird doing it in toml file

LGonzalezGomez:
Version can be found and downloaded here
https://test.pypi.org/project/cluster-experiments/0.20.1/

Must still fix GHA to use uv

.github/workflows/docs.yml
```diff
@@ -16,5 +16,6 @@ jobs:
       - run: cp README.md docs/index.md
       - run: cp -r theme docs/theme
       - run: make install-dev
+      - run: source .venv/bin/activate
```
david26694, line 19: if we don't do this, does it not work?

setup.py
```diff
@@ -1,71 +1,5 @@
 from setuptools import find_packages, setup
```
david26694, line 1: do we need a setup.py now?

david26694 (commented):
Couple comments, but apart from that, let's just change the readme to require uv for contributing and merge

Makefile
david26694: not really

david26694 (approved):

Issue #182, opened 2024-07-04
Allow relative perturbator with mixed-sign values


Gabrielcidral1:
In some cases, you might want to add a relative perturbator for a target that includes negative and positive values, such as profit margin. In this case, with a positive effect of 0.5, and a target of -100,  we would the result to be -50.

@david26694 Should we create a new class for it? 

Here's a example (suggested by @matheusbuniotto)

```
class RelativeConstantPerturbator(Perturbator):
    def perturbate(
        self, df: pd.DataFrame, average_effect: Optional[float] = None
    ) -> pd.DataFrame:
        """
        Usage:

        ```python
        from cluster_experiments.perturbator import UniformPerturbator
        import pandas as pd
        df = pd.DataFrame({"target": [1, 2, 3], "treatment": ["A", "B", "A"]})
        perturbator = UniformPerturbator()
        perturbator.perturbate(df, average_effect=1)
        ```
        """
        df = df.copy().reset_index(drop=True)
        average_effect = self.get_average_effect(average_effect)
        df = self.apply_multiplicative_effect(df, average_effect)
        return df

    
    def apply_multiplicative_effect(
        self, df: pd.DataFrame, effect: Union[float, np.ndarray]
    ) -> pd.DataFrame:
        mask = df[self.treatment_col] == self.treatment
        original_values = df.loc[mask, self.target_col]
        
        # Calculate the new values
        new_values = np.where(
            original_values >= 0,
            original_values * (1 + effect),
            original_values * (1 - effect)
        )
        
        # Update the dataframe
        df.loc[mask, self.target_col] = new_values
        
        return df
```

david26694:
Hey gabi, thanks for the issue. Couple questions?
* Any use-case or example you can think this is covering?
* I understand if effect is 0.5 and target is [-100, 10], then perturbated values are [-50, 15], right?

Gabrielcidral1:
Our use case is to perturbate effect on profit margin, where it can be either positive or negative. And yes, your example is what we have in mind.

david26694:
yep, feel free to create a PR!

PR #196, opened 2024-11-22
feat: add relative mixed perturbator

Gabrielcidral1:
As per this [issue](https://github.com/david26694/cluster-experiments/issues/182), this PR adds a new type of perturbator to allow a target with mixed signals (positive and negative.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -276,6 +276,7 @@ The library offers the following classes:
     * `ConstantPerturbator`: to artificially perturb treated group with constant perturbations
     * `BinaryPerturbator`: to artificially perturb treated group for binary outcomes
     * `RelativePositivePerturbator`: to artificially perturb treated group with relative positive perturbations
+    * `RelativeMixedPerturbator`: to artificially perturb treated group with relative perturbations for positive and negative targets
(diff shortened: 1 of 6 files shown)


Gabrielcidral1:
@david26694 shouldn't we block automatically merges to main?

david26694 (changes requested):
Good! It's missing from:
* test_docs
* README
* init

Also need a library bump. Aside from that, lgtm!

david26694:
> @david26694 shouldn't we block automatically merges to main?

I think it is blocked
imagen

gabrielristow:
With this implementation, the RelativePositivePerturbator becomes a bit redundant,right? But probably we don't want to deprecate it.

gabrielristow:
> Good! It's missing from:
> 
> * test_docs
> * README
> * init
> 
> Also need a library bump. Aside from that, lgtm!

Sorry forgot about those, it's done!

Gabrielcidral1:
> > @david26694 shouldn't we block automatically merges to main?
> 
> I think it is blocked imagen

Take a look 
https://github.com/david26694/cluster-experiments/pull/199

david26694 (approved):

david26694:
great!

PR #208, opened 2024-12-15
feat: add Delta Method Analysis

LGonzalezGomez:
This PR adds Delta Method Analysis. It works with PowerAnalysis pipelines out-of-the-box.
Also added a couple of unit tests to ensure it is working appropiately.

It still needs to be hooked up with the Inference pipeline. Expected to do it at a later stage to split the PR.


cluster_experiments/power_config.py
```diff
@@ -254,6 +255,7 @@ def _raise_error_if_missing(self, attr, other_attr):
     "ttest_clustered": TTestClusteredAnalysis,
     "paired_ttest_clustered": PairedTTestClusteredAnalysis,
     "mlm": MLMExperimentAnalysis,
+    "delta": DeltaMethodAnalysis,
```
david26694, line 258: in the notebook, there's a comment about GEE. It needs to be removed

cluster_experiments/experiment_analysis.py
```diff
@@ -1204,3 +1206,169 @@ def _split_pre_experiment_df(self, df: pd.DataFrame):
...
+
+class DeltaMethodAnalysis(ExperimentAnalysis):
+    def __init__(
+        self,
+        cluster_cols: Optional[List[str]] = None,
+        target_col: str = "target",
+        scale_col: str = "scale",
+        treatment_col: str = "treatment",
+        treatment: str = "B",
+        covariates: Optional[List[str]] = None,
```
david26694, line 1219: let's drop all the fields we're not currently using

cluster_experiments/experiment_analysis.py
```diff
@@ -1204,3 +1206,169 @@ def _split_pre_experiment_df(self, df: pd.DataFrame):
...
+
+        group_mean = target_sum / scale_sum
+        group_variance = (
+            (1 / (scale_mean**2)) * target_variance
+            + (target_mean**2) / (scale_mean**4) * scale_variance
+            - (2 * target_mean) / (scale_mean**3) * target_scale_cov
+        ) / group_size
+        return group_mean, group_variance
+
+    def _get_mean_SE(self, df: pd.DataFrame) -> tuple[float, float]:
```
david26694, line 1316: can we use standard_error instead of SE?

cluster_experiments/power_config.py
david26694, line 258: also, you're adding it here but not adding the scale parameter in here. I think we need to add it if we want to init via config. I would also test init via config, I think it's missing

README.md
```diff
@@ -263,6 +263,7 @@ The library offers the following classes:
     * `PairedTTestClusteredAnalysis`: to run a paired t-test on aggregated data for clusters
     * `ClusteredOLSAnalysis`: to run OLS analysis on the results of a clustered design
     * `OLSAnalysis`: to run OLS analysis for non-clustered data
+    * `DeltaMethodAnalysis`: to run Delta Method Analysis on clustered design
```
david26694, line 266: *for clustered design, I'd say

david26694: or, better:
for clustered designs

cluster_experiments/experiment_analysis.py
david26694: (at least ratio_covariates)

tests/analysis/test_analysis.py
```diff
@@ -251,3 +256,49 @@ def test_get_inference_results(experiment_analysis, analysis_df_diff):
...
+        p_values.append(analyser.get_pvalue(data))
+
+    positive_rate = sum(p < 0.05 for p in p_values) / len(p_values)
+
+    assert positive_rate == pytest.approx(
+        0.05, abs=0.01
+    ), "P-value A/A calculation is incorrect"
+
+
+def test_stats_delta_vs_ols(analysis_ratio_df, experiment_dates):
```
david26694, line 286: I'd also test that you get the same results with aggregated and non-aggregated data

tests/analysis/test_analysis.py
```diff
@@ -251,3 +256,49 @@ def test_get_inference_results(experiment_analysis, analysis_df_diff):
...
+
+    assert 0.05 >= analyser.get_pvalue(df_experiment) >= 0
+
+
+def test_aa_delta_analysis(dates):
+
+    analyser = DeltaMethodAnalysis(cluster_cols=["user"], scale_col="scale")
+
+    p_values = []
+    for _ in range(1000):
```
david26694, line 275: wdyt about seeding these tests?

david26694 (changes requested):
lgtm but suggesting a few minor changes. Also bump the library via the toml, please :)

david26694:
Also, create issues for:
* Inference pipeline
* Variance reudction

cluster_experiments/power_config.py
LGonzalezGomez: I am not sure what you mean here. I get that it might be tricky to know that you have to pass an additional parameter or that the "scale" column has to appear somewhere. Is this what you are refering to?
If that is the case, do you have any recommendation on how to go around this? I am not sure I understand how to do it properly...

tests/analysis/test_analysis.py
david26694: yep, be4-hand

david26694: maybe it's not needed, I just think we need to test the aggregation in some way (I think we're not testing it), wdyt is best?

cluster_experiments/power_config.py
LGonzalezGomez: It was pretty clear, thanks!
I think that now it should be working. I had to do some changes and raise a couple of additional errors to save guard. 
When time allows, could you check again and give more feedback?
Thanks!

PR #219, opened 2025-01-14
add analysis plan from config

david26694:
Allow instantion of analysis plan from metrics config. Need to create another PR to instantiate from "pure" config.

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,20 +136,25 @@ variants = [
 city_dimension = Dimension(name="city", values=["NYC", "LA"])
 
 # Create analysis plan
-plan = AnalysisPlan.from_metrics(
-    metrics=[aov, delivery_time],
-    variants=variants,
-    variant_col="experiment_group",
-    dimensions=[city_dimension],
-    analysis_type="clustered_ols",
-    analysis_config={
-        "cluster_cols": ["customer_id"],
-    },
-)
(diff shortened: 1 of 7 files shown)


ludovico-lanni (commented):
Looks good! What about instantiating analysis plan via multiple hypothesis tests instead of via the shortcut of using the from_metrics builder?

What I mean is that this config-based analysis plan creation covers the from_metrics() shortcut, and we could also cover the more flexible and verbose way to instantiate hypothesis tests via config, and in turns also the analysis plan via a list of hypothesis tests configs. A quite frequent example use case would be to analyse 5 metrics via OLS where each of them requires a specific covariate. A single analysis plan from_metrics is not possible as each hypothesis test requires a different analysis_config dict (due to the different covariate)

david26694:
yep, this is what I meant by this:
> Need to create another PR to instantiate from "pure" config.
Essentially we've writen from_metrics_dict and we need to write another from_dict, which will cover your case I think. Wdyt?

ludovico-lanni:
> Essentially we've writen from_metrics_dict and we need to write another from_dict, which will cover your case I think. Wdyt?

I see! It makes sense to me :)

ludovico-lanni (approved):
lgtm

PR #230, opened 2025-03-02
feat: Add Variance Reduction to Delta Method and couple it with CUPAC

LGonzalezGomez:
CUPED is used with the Delta Method. 
This is intended to be used directly with the Power Analysis so the user is expected to use CUPAC. This simplifies a bit the usage although it is a slight approximation but should work for our implementation. 
Otherwise we have to change how the interface works so we can pass ratio_metrics and normal metrics, which might imply a ton of refactoring and changing how users usually interact with the library.

Please feel free to check it and we can discuss implementation and align on the best way going forward!

image
image
Screenshot 2025-03-02 at 20 00 08

### DIFF
--- a/cluster_experiments/experiment_analysis.py
+++ b/cluster_experiments/experiment_analysis.py
@@ -1225,7 +1225,6 @@ class DeltaMethodAnalysis(ExperimentAnalysis):
         treatment_col: str = "treatment",
         treatment: str = "B",
         covariates: Optional[List[str]] = None,
-        ratio_covariates: Optional[List[Tuple[str]]] = None,
         hypothesis: str = "two-sided",
     ):
         """
@@ -1239,7 +1238,6 @@ class DeltaMethodAnalysis(ExperimentAnalysis):
             treatment_col: name of the column containing the treatment variable.
             treatment: name of the treatment to use as the treated group.
             covariates: list of columns to use as covariates.
-            ratio_covariates: list of tuples of columns to use as covariates for ratio metrics. First element is the numerator column, second element is the denominator column.
             hypothesis: one of "two-sided", "less", "greater" indicating the alternative hypothesis.
 
             Usage:
@@ -1274,7 +1272,6 @@ class DeltaMethodAnalysis(ExperimentAnalysis):
         self.scale_col = scale_col
         self.cluster_cols = cluster_cols or []
         self.covariates = covariates or []
-        self.ratio_covariates = ratio_covariates or []
 
         if cluster_cols is None:
             raise ValueError(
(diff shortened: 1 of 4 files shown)


LGonzalezGomez:
Codecov is not updating. The only line that is not covered is present in CUPAC. Not sure if it is really needed, seems a bit too much IMO, but we could add a test in power config to test that CUPAC with Delta Method is indeed working.

cluster_experiments/experiment_analysis.py
```diff
@@ -1301,30 +1365,23 @@ def _aggregate_to_cluster(self, df: pd.DataFrame) -> pd.DataFrame:
...
-        target_mean, scale_mean = df.loc[:, [self.target_col, self.scale_col]].mean()
-        target_variance, scale_variance = df.loc[
-            :, [self.target_col, self.scale_col]
-        ].var()
-        target_sum, scale_sum = df.loc[:, [self.target_col, self.scale_col]].sum()
-
-        target_scale_cov = df.loc[:, self.target_col].cov(df.loc[:, self.scale_col])
+        df = df.copy()
+        df = self._transform_metrics(df)
+        if self.covariates:
```
david26694, line 1375: wondering if we could avoid this if else in any way

cluster_experiments/experiment_analysis.py
```diff
@@ -1333,6 +1390,14 @@ def _get_mean_standard_error(self, df: pd.DataFrame) -> tuple[float, float]:
         Variance reduction is used if covariates are given.
         """
 
+        if (self._get_num_clusters(df) < 1000).any():
```
david26694, line 1393: could the 1000 be a class variable?

cluster_experiments/experiment_analysis.py
```diff
@@ -1333,6 +1390,14 @@ def _get_mean_standard_error(self, df: pd.DataFrame) -> tuple[float, float]:
...
+            self.__warn_small_group_size()
+
+        if self.covariates:
```
david26694, line 1396: why are covariates relevant to aggregation?

cluster_experiments/experiment_analysis.py
```diff
@@ -1457,6 +1522,24 @@ def from_config(cls, config):
...
+            raise ValueError(
+                "The data should be aggregated at the cluster level for the Delta Method analysis using CUPED."
+            )
+
+    def _get_num_clusters(self, df):
+        """
+        Check if there are enough clusters to run the analysis.
+        """
+        return df.groupby(self.treatment_col).apply(
+            lambda x: self._get_cluster_column(x).nunique()
```
david26694, line 1540: why do we groupby treatment col?

tests/analysis/test_analysis.py
```diff
@@ -324,8 +325,124 @@ def test_stats_delta_vs_ols(analysis_ratio_df, experiment_dates):
     SE_delta = analyser_delta.get_standard_error(df)
 
     assert point_estimate_delta == pytest.approx(
-        point_estimate_ols, rel=1e-2
+        point_estimate_ols, rel=1e-1
```
david26694, line 328: could you add a test where the Deltamethod is passed to an analysisPlan?

david26694 (approved):

Issue #229, opened 2025-02-14
Add logic for aggregations in PowerAnalysis


david26694:
Let's say I have a metric that, for each unit, changes as the experiment grows, like a cumulative metric in https://arxiv.org/abs/2411.06150, or number of orders per user. We need to add an aggregation layer in the PowerAnalysis class

PR #246, opened 2025-10-10
Add Sliding MDE analysis (solved issue #229)

ludovicolc:
Add mde_sliding_time_line method in the NormalPowerAnalysis class, enabling the calculation of MDE over varying experiment lengths.

Key features:

- Computes MDE across sliding time windows, allowing measurement of treatment impact on metrics that vary over time.
- Supports custom aggregation functions (agg_func) for grouped metrics (e.g., sum, mean, median).
- Allows element-wise post-processing (post_process_func) on the aggregated metric, useful for transforming counts into binary flags or applying other transformations.

This enhancement provides flexibility to analyze temporal patterns in metric response and better capture dynamics of retention, conversion, or other evolving metrics.

Issue #229


ludovicolc:
added notebook to normal_power_lines.ipynb (same file as power_time_line and mde_time_line)

PR #264, opened 2026-03-20
feat:  Relative Effects and MDE for Ratio Metrics

luizhsuperti:
### Summary
This PR adds proper relative (percent) lift estimation and minimum detectable effect (MDE) calculation for ratio metrics using an outer delta method, instead of relying on naive scaling by the control mean.

### Problem
For ratio metrics such as conversion rate or revenue per user:

- Naive relative lift (absolute effect divided by control mean) underestimates uncertainty because the denominator is treated as fixed.
- Naive relative MDE from scaling absolute MDE is only an approximation and can be inaccurate when denominator variance matters.

### What Changed

- Added DeltaMethodLiftTransformer with:
  - fit(mean_diff, var_abs, ctrl_mean, ctrl_var)
  - lift_and_se(...) static helper for relative lift and SE
  - relative_mde(alpha, power, ctrl_mean, ctrl_var, treat_var) static helper solving the quadratic power equation
- Extended DeltaMethodAnalysis with relative_effect (default false):
  - When true, uses DeltaMethodLiftTransformer internally to return relative effect and proper SE
  - Works with and without covariates (including CUPED flows)
- Updated PowerAnalysis config validation to allow relative_effect with DeltaMethodAnalysis
- Added documentation notebook showing:
  - naive vs proper relative lift
  - naive vs proper relative MDE
  - integration with NormalPowerAnalysis and AnalysisPlan
- Added tests for:
  - transformer math
  - DeltaMethodAnalysis integration
  - config behavior
  - end-to-end AnalysisPlan and power workflows

### Impact
Users can now estimate relative effects for ratio metrics with uncertainty propagation that is statistically consistent with delta-method assumptions, and compute a principled relative MDE via the quadratic formulation.

### Notes
This keeps backward compatibility while introducing the new transformer API and helper functions for direct use in custom workflows.


cluster_experiments/experiment_analysis.py
```diff
@@ -1784,6 +1787,19 @@ def _get_mean_standard_error(self, df: pd.DataFrame) -> tuple[float, float]:
         mean_diff = treat_mean - ctrl_mean
         standard_error = np.sqrt(treat_var + ctrl_var)
 
+        if self.relative_effect:
+            transformer = DeltaMethodLiftTransformer(self.treatment_col)
+            transformer.fit(
+                mean_diff=mean_diff,
+                var_abs=treat_var + ctrl_var,
```
david26694: perhaps you can pass std_error instead of var,and you have it in this variable: standard_error

cluster_experiments/experiment_analysis.py
```diff
@@ -1784,6 +1787,19 @@ def _get_mean_standard_error(self, df: pd.DataFrame) -> tuple[float, float]:
...
+                ctrl_mean=ctrl_mean,
+                ctrl_var=ctrl_var,
+            )
+            return (
```
david26694: for compatibility, I would not return anything on fit, just store values in attributes

cluster_experiments/relative_lift_transformer.py
```diff
@@ -152,3 +152,149 @@ def summary(self):
...
+            index=[self.treatment_col],
+            columns=[0, 1],
+        )
+
+    def summary(self):
+        return {
+            "percent_lift": self._relative_lift_value,
+            "_se_relative_lift": self._se_relative_lift,
+            "pvalue": self.pvalues[self.treatment_col],
+            "conf_int": self.conf_int(0.05).loc[self.treatment_col],
```
david26694: shouldnt we use alpha in here?

david26694 (changes requested):

tests/analysis/test_delta_relative.py
```diff
@@ -0,0 +1,649 @@
...
+    ols_se = ols_rel.get_standard_error(df_agg)
+    delta_se = delta_rel.get_standard_error(df_agg)
+
+    # Both should detect a positive effect in the same direction
+    assert ols_point > 0
+    assert delta_point > 0
+
+    # Point estimates come from different estimators (unweighted vs weighted by
+    # scale) so a 5% relative tolerance is appropriate
+    assert ols_point == pytest.approx(delta_point, rel=0.05)
```
david26694, line 993: just curious. do you think we can get it closer?

cluster_experiments/relative_lift_transformer.py
```diff
@@ -117,38 +167,113 @@ def fit(
...
+        Minimum detectable relative lift (two-sided, double-delta quadratic).
 
-    def summary(self):
-        return {
-            "percent_lift": self._relative_lift_value,
-            "_se_relative_lift": self._se_relative_lift,
-            "pvalue": self.pvalues[self.treatment_col],
-            "conf_int": self.conf_int(0.05).loc[self.treatment_col],
-        }
+        Solves A*m^2 + B*m + C = 0 for the smallest positive m satisfying
```
david26694: this is very new to me, could you add some reference or small proof?

david26694 (changes requested):

cluster_experiments/relative_lift_transformer.py
```diff
@@ -21,12 +21,62 @@ def conf_int(self, alpha: float) -> Dict[str, Tuple[float, float]]: ...
...
+            index=[self.treatment_col],
+            columns=[0, 1],
+        )
+
+    def summary(self):
+        return {
+            "percent_lift": self._relative_lift_value,
+            "_se_relative_lift": self._se_relative_lift,
+            "pvalue": self.pvalues[self.treatment_col],
+            "conf_int": self.conf_int(0.05).loc[self.treatment_col],
```
david26694: let's change this 0.05 to alpha

tests/analysis/test_delta_relative.py
```diff
@@ -0,0 +1,649 @@
...
+    transformer = DeltaMethodLiftTransformer("treatment")
+    transformer.fit(
+        mean_diff=mean_diff,
+        std_error=np.sqrt(var_abs),
+        ctrl_mean=ctrl_mean,
+        ctrl_var=ctrl_var,
+    )
+
+    assert transformer.bse["treatment"] >= se_naive
+    assert transformer.bse["treatment"] == pytest.approx(se_naive, rel=0.15)
```
david26694, line 163: this also looks big

david26694 (commented):
.

tests/analysis/test_delta_relative.py
```diff
@@ -0,0 +1,649 @@
...
+
+    def mock_ppf(q):
+        # z_alpha for q=0.975 and z_beta for q=0.8
+        return 1.96 if q > 0.9 else 2.0
+
+    monkeypatch.setattr(
+        "cluster_experiments.relative_lift_transformer.stats.norm.ppf", mock_ppf
+    )
+
+    with pytest.raises(ValueError, match="invalid power equation"):
```
david26694: in which scenario do we not get an MDE?

cluster_experiments/relative_lift_transformer.py
```diff
@@ -117,38 +167,113 @@ def fit(
...
+        relative_lift = mean_diff / ctrl_mean
+        var_relative = (
+            var_abs / (ctrl_mean**2)
+            + (mean_diff**2) * ctrl_var / (ctrl_mean**4)
+            + 2 * mean_diff * ctrl_var / (ctrl_mean**3)
         )
+        return relative_lift, float(np.sqrt(var_relative))
+
+    @staticmethod
+    def relative_mde(
```
david26694: after having a look I like this, but I think at some point it better bolngs in the power analysis class. Maybe we can merge this changing to internal method but later we move to NormalPowerAnalysis and use it to calculate the relative mde of ols
```suggestion
    def _relative_mde(
```

david26694: but I don't see this method being called anyway, is this correct?

david26694 (changes requested):

cluster_experiments/relative_lift_transformer.py
luizhsuperti: oki! I'll add references in the .py, but I'll add a more thorough view in a notebook

tests/analysis/test_delta_relative.py
luizhsuperti: I'm adding scenarios hashtags comments, but be more specific on the types of errors where the MDE fails (ie, high variance, imcompatible variance with asked alpha,etc)

tests/analysis/test_delta_relative.py
luizhsuperti: hmm  did not think of that. maybe some simulations to check how close, or rather will check in some other sources

cluster_experiments/relative_lift_transformer.py
```diff
@@ -117,38 +167,150 @@ def fit(
...
+        relative_lift = mean_diff / ctrl_mean
+        var_relative = (
+            var_abs / (ctrl_mean**2)
+            + (mean_diff**2) * ctrl_var / (ctrl_mean**4)
+            + 2 * mean_diff * ctrl_var / (ctrl_mean**3)
         )
+        return relative_lift, float(np.sqrt(var_relative))
+
+    @staticmethod
+    def _relative_mde(
```
david26694: imo this is a responsibility of the PowerAnalysis class, I would move it in there

cluster_experiments/experiment_analysis.py
```diff
@@ -1784,6 +1787,19 @@ def _get_mean_standard_error(self, df: pd.DataFrame) -> tuple[float, float]:
         mean_diff = treat_mean - ctrl_mean
         standard_error = np.sqrt(treat_var + ctrl_var)
 
+        if self.relative_effect:
+            transformer = DeltaMethodLiftTransformer(self.treatment_col)
```
david26694, line 1894: a bit confused in here, where is this method used? does this handle covariates?

luizhsuperti: Used by [_get_mean_standard_error] (p-value/SE path) and analysis_standard_error_with_stats(power path).
For Covariates: yes. It calls [_get_group_statistics] which builds [thetas_dict] from [self.covariates] and applies CUPED variance reduction inside [_get_group_mean_and_variance] So relative effects respect covariates.

david26694:
closing this, new PR: https://github.com/david26694/cluster-experiments/pull/269

PR #269, opened 2026-08-17
Mde new ratio metrics

david26694:
Adds relative-effect support for ratio metrics via the delta method, and reworks power analysis so power and the MDE are derived from the same object.

## The core change

For an absolute effect, the standard error does not depend on the effect size. For a **relative** effect it does, because the baseline in the denominator of the lift is itself estimated. Both estimators that support relative effects share the same variance shape, now captured by `StandardErrorCurve`:

```
SE(m)**2 = SE(0)**2 + effect_var * m**2 - 2 * effect_cov * m
```

The delta method is the special case `effect_cov == -effect_var` (independent arms), which collapses to `se2_t + se2_c * (1 + m)**2`.

Power and the MDE are both derived from that curve, which is what makes them exact inverses. Previously the MDE used an effect-dependent formula while power used a constant standard error, so they disagreed: `mde(power=0.8)` returned an effect whose `power_line` was `0.809`.

The power condition solved is `m = (z_alpha + z_beta) * SE(m)`, per the project slides on relative lift for OLS.

## Breaking changes

Two, both intentional:

1. **`mde_rolling_time_line` no longer returns the `relative_mde` key.** It was `mde / mean(target)` — a second, systematically optimistic notion of relative MDE that treats the baseline as a fixed constant. Relative analyses had both normalisations applied at once. There is now one way to ask for a relative MDE: `relative_effect=True`. Migration for callers that want the old number: `df["mde"] / baseline`.

2. **`run_average_standard_error` was removed.** It had no callers.

Also changed, not breaking: `OLSAnalysis` and `ClusteredOLSAnalysis` with `relative_effect=True` now return **larger** MDEs, because they get the same effect-dependent treatment as the delta method instead of the linear approximation. `test_relative_ols_vs_delta_parity` already asserted the two estimators agree; their MDEs now do too.

## Statistical caveat

The convention shipped here is **conservative** relative to the exact Wald rejection boundary, which is the fixed point of `x = z_alpha * SE(x)` rather than `z_alpha * SE(m)`. It never under-sizes an experiment. The gap is 0.5% at 2% baseline relative standard error and 3% at 5%, growing to 1.5x at 20% — a regime where the delta-method interval is not trustworthy anyway (this is Fieller's problem).

`mde_threshold_derivation.md` has the derivation, the ordering proof, and the numbers. That file is currently at the repo root and should probably move under `docs/` or be dropped before merge.

## Absolute effects are unchanged

`tests/power_analysis/test_absolute_effect_golden.py` pins power and MDE for `OLSAnalysis`, `ClusteredOLSAnalysis` and `DeltaMethodAnalysis` across all three hypotheses and all five public power/MDE methods, against values captured from `main`. 9 cases, `rel=1e-12`.

## Also included

A fix to the delta-method CUPED covariate mean, which was computed as the covariate sum over the scale sum instead of the scale-weighted mean. The resulting offset cancels out of the treatment-control difference, so absolute effects were unaffected and the bug was invisible — but it lands in the denominator of a relative lift. On a planted +10% lift the relative estimate came back as +193.6%. The bug predates this branch (it arrived with the CUPED-for-delta feature in `bbc2092` and is on `main`); `relative_effect` is only what exposes it.

## Tests

`tests/analysis/test_delta_relative.py` went from 42 test functions to 22, chosen by mutation testing rather than by reading: twelve bugs were injected into the relative-lift and MDE code, recording which tests caught which. Every mutation is still caught, and no two remaining tests share a mutation signature.

Removed for being unable to fail — one asserted `(x*c)/c == x`, one derived its reference from the value under test (which is why it passed throughout the CUPED bug), one passed with the quadratic solver removed — and for duplication, where the same identity `m/SE(m) == z_alpha + z_beta` was asserted by three separate tests.

The delta-versus-OLS comparison was also replaced. It compared the delta method on `target`/`scale` against OLS on a precomputed per-cluster `ratio` column, which are different estimands: one weights clusters by scale, the other does not. Both now receive the same **un-aggregated** frame, one row per observation, cluster sizes 1 to 24. OLS over raw rows is the difference in row-level means, which is the ratio metric; the delta method reaches it by aggregating to `(sum, count)`. Neither side is reweighted by hand, and they agree on the relative lift to `1e-9`, the standard error to 0.07%, and the MDE to under 1%.

Relative OLS gained coverage it did not have: curve coefficients against an independently computed regression covariance, MDE and power inverting exactly, and MDE above the linear approximation. It is the estimator whose MDE changed value on this branch and it previously had no MDE tests. Also added a negative control mean, which nothing reached before.

## Notes

- Full suite: 620 passing
- `docs/relative_delta.ipynb` is registered in `mkdocs.yml`
- No version bump; this repo bumps in separate release commits

🤖 Generated with [Claude Code](https://claude.com/claude-code)


cluster_experiments/power_analysis.py
```diff
@@ -856,20 +857,30 @@ def _get_standard_error(
...
-            yield self.analysis.get_standard_error(split_df)
+            yield self.analysis.get_standard_error_curve(split_df)
 
     def _normal_power_calculation(
-        self, alpha: float, std_error: float, average_effect: float
+        self, alpha: float, se_curve: StandardErrorCurve, average_effect: float
     ) -> float:
         """Returns the power of the analysis using the normal distribution.
+
+        The standard error is evaluated at the effect being tested. For an
```
david26694: remove comment

luizhsuperti:
@david26694 I went through the rework and did a verification locally:
>
> - **Full suite on `ae0287c`: 594 passed, 0 failed.**
> - **Golden values independently reproduced**: I checked out `main`, recomputed all 135 power/MDE numbers with the harness from `test_absolute_effect_golden.py`, and compared against the `EXPECTED` dict — bit-identical (worst relative diff 0.0). The claim that absolute effects are untouched holds exactly.
> - **CUPED fix (#270)**: verified the scale-weighted covariate mean is what `_correct_target` requires — the old unweighted mean cancels in absolute differences but lands in the denominator of a ratio, which explains the +193.6% artifact.
> - **Test cut 42→22**: the mutation-testing rationale checks out; the deleted tests I spot-checked were tautological or self-referential (which is exactly why the CUPED bug survived them).
> - The `StandardErrorCurve` design resolves the MDE/power inconsistency by construction, and the de-mystified solver in `ae0287c` is now readable end-to-end. This is a much better shape than my original approach — thanks for the thorough rework.


We should open the follow-up issues flagged: (derivation docs page, units-on-results, `relative_effect` whitelist).

- Units on results — power/MDE/point-estimate floats don't carry their scale; the same method returns absolute units or relative lift depending on the relative_effect constructor flag. Consider a small result type or unit-suffixed keys so results are self-describing.
- relative_effect capability check — PowerConfig hardcodes {OLSAnalysis, ClusteredOLSAnalysis, DeltaMethodAnalysis}; move this to a capability declared by the analysis class itself so direct construction is also guarded and subclasses aren't locked out.
- Derivation docs page — mde_threshold_derivation.md was removed in dc4d414 in favour of inline comments; consider a docs-site page for the SE-curve/MDE quadratic derivation, registered in mkdocs.yml nav.

luizhsuperti:
@copilot+claude-opus-4.8:max double check the comment above and general PR

